PR #48, opened 2014-04-11
Removal of Mono.Data.Sqlite dependency

fogzot:
The changes in this pull request don't change the necessity to have SQlite (and Mono.Data.Sqlite) available at runtime to use DefaultSecureMimeContext but remove the hard dependency on Mono.Data.Sqlite DLL. If you agree that this is good we can discuss how to really make IX509CertificateStore pluggable.


mwpowellhtx:
Will work for now, thank you. If I discover any different will follow up. Thanks! :+1:

PR #117, opened 2015-03-20
Add immutability-friendly ctors to MimeMessage

### LINKED ISSUE #116: Add more immutability-friendly APIs
When using MimeKit & MailKit from F#, some things look weird, because of how F# programs you to think in immutability all around.

For example, I would prefer that MimeMessage ctor not be parameterless, and accept subject, body, and arrays for From & To parameters (to avoid using the setters of properties).

This way an F# developer could use this API without the need of using the *evil* assignment operator that we're now allergic to :-)

Would you be willing to add such a constructor? If the answer is "Yes, patches welcome", I'll happily do it when I've some time in my hands.

Thanks

knocte:
Fixes #116


knocte:
@jstedfast Updated PR to address your comments.

PR #143, opened 2015-06-09
Fix ChainedStream to handle partial reads

jdasilva:
I was working with MimeKit to read an HttpWebResponse and found that the content was not being fully read.  Apparently Stream.Read can return fewer bytes than requested and the response stream will.  This fixes ChainedStream to continue reading until the entire expected read is fulfilled.  An alternative solution would be to pass partial reads through.  

I also modified the ChainedStream tests to utilize a stream that always partial reads.  This causes the last three tests to fail and the fix corrects them.  I wasn't sure if this would be better a separate test or not.  I can also confirm that this did fix my issue with HttpWebResponse.

Let me know if there is anything else I can do to make this a better fit for the project.

### DIFF
--- a/MimeKit/IO/ChainedStream.cs
+++ b/MimeKit/IO/ChainedStream.cs
@@ -281,14 +281,15 @@ namespace MimeKit.IO {
 			int n, nread = 0;
 
 			while (current < streams.Count) {
-				if ((n = streams[current].Read (buffer, offset + nread, count - nread)) > 0) {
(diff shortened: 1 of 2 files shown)


jstedfast:
I ended up fixing it a slightly different way, but I added your unit test. Thanks for the bug report and patch!

jdasilva:
Great - and thank you for all the work on MimeKit!

PR #216, opened 2016-01-03
Adds support additional private key files

matbech:
This adds support reading additional private key files. For example, when loading PEM files from openssl PemReader returns an RSAPrivateCrtKeyParameters object. This object implements AsymmetricKeyParameter and not AsymmetricCipherKeyPair. With the proposed changes both AsymmetricCipherKeyPair and AsymmetricKeyParameter derived objects are supported.


jstedfast:
Thanks. I've committed a slightly different patch for this.

matbech:
Thanks

PR #271, opened 2016-10-14
Add Switch to HtmlToHtml converter to strip comments from output.

DanDude0:
Set 'StripComments' to true to remove comments from the output HTML.


jstedfast:
Ended up implementing this as FilterComments instead of StripComments.

PR #283, opened 2017-03-02
Add support for CryptographyContext factories

dusk0r:
Currently there is no way to control the creation of (S/MIME / OpenPGP) CryptographyContexts. I propose optional factory methods.

### DIFF
--- a/MimeKit/Cryptography/CryptographyContext.cs
+++ b/MimeKit/Cryptography/CryptographyContext.cs
@@ -41,9 +41,9 @@ namespace MimeKit.Cryptography {
 	public abstract class CryptographyContext : IDisposable
 	{
 		const string SubclassAndRegisterFormat = "You need to subclass {0} and then register it with MimeKit.Cryptography.CryptographyContext.Register().";
-		static ConstructorInfo SecureMimeContextConstructor;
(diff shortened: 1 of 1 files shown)


jstedfast:
Why not just subclass the context you want and register it's Type?

jstedfast:
I much prefer the way it is now...

dusk0r:
Because you can't use non-static external resources.

jstedfast:
Can you give me an example of what you are trying to accomplish that can't be done by subclassing?

jstedfast:
Keep in mind that your PR breaks existing users of MimeKit... :(

jstedfast:
Actually, n/m, I looked at your PR again and it doesn't...

dusk0r:
We have several ASP.NET webapps. On startup they fetch a private S/MIME Key to sign outgoing mails. I don't want to embed or store the certificate on the machine.

Right now my best solution is to fetch the key and store it into a static property. Then subclass the context and load the key from that property. But I don't like that solution.

jstedfast:
Since it doesn't break API/ABI, I'm willing to merge this, but can you fix the code formatting/indenting to match the rest of the code? Thanks.

jstedfast:
Awesome, thank you.

PR #298, opened 2017-04-14
Place break inside if block

jvitkauskas:
Most likely fixed a bug


jstedfast:
Looks reasonable but I don't really know the code. I just imported Mono.Data.Sqlite from the [Mono](https://github.com/mono/mono) project for use with my unit tests.

I just did a comparison and it looks like some changes have gone in since I last imported the code but it doesn't look like Mono has your fix.

Would you like to submit your fix there? You'll find the SQLiteConnection.cs file in `mcs/class/Mono.Data.Sqlite/Mono.Data.Sqlite_2.0/`

jvitkauskas:
Ok, opened a PR in Mono https://github.com/mono/mono/pull/4697

jstedfast:
Cool, thanks. I'll merge this in MimeKit as well.

PR #333, opened 2017-08-29
Don't set TextBody to contents of text attach when body undefined

AntGraham:
Hi,

When the following mail is parsed:
```
From: sender@domain.com
Date: Tue, 29 Aug 2017 09:45:39 +1000
Subject: This has no body, just a text attachment
Message-Id: <75SXBEJJ72U4.5KFFZ6J56L2T2@localhost.localdomain>
MIME-Version: 1.0
Content-Type: text/plain; name="Plain Text.txt"
Content-Disposition: attachment; filename="Plain Text.txt"
Content-Transfer-Encoding: 7bit

This is the text attachment
```
The Message.TextBody is set to the contents of the attachment, rather than being null as expected.
This is because the message doesn't have any body.  The fix is just to add a `&& !body.IsAttachment` to the condition in `MimeMessage.GetTextBody` on line 924.

I don't believe the RFC states you have to have a body, and email clients render this as expected - a blank message with a small text attachment.

Many thanks!

Ant.

### DIFF
--- a/MimeKit/MimeMessage.cs
+++ b/MimeKit/MimeMessage.cs
@@ -921,7 +921,7 @@ namespace MimeKit {
 			} else {
 				var body = Body as TextPart;
 
-				if (body != null && body.IsFormat (format))
+				if (body != null && body.IsFormat (format) && !body.IsAttachment)
 					return body.Text;
 			}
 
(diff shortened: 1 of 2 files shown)


AntGraham:
These changes are as minimal as possible.

If the one attachment part shouldn't be interpreted as a body as I am suggesting, another fix is to not set it as the Body property if 'IsAttachment is true', rather than checking the Body property when retrieving the text property as has been done here.

jstedfast:
Thanks. Looks like I had that logic for text parts inside multiparts but not for the case you fixed.

PR #337, opened 2017-09-12
Fix RC2CBC Parameter parsing.

alxbauer:
Parameters are encoded as Sequence. See: https://tools.ietf.org/html/rfc3370#section-5.2

### DIFF
--- a/MimeKit/Cryptography/SecureMimeContext.cs
+++ b/MimeKit/Cryptography/SecureMimeContext.cs
@@ -833,17 +833,20 @@ namespace MimeKit.Cryptography {
 			}
 
 			if (identifier.Algorithm.Id == CmsEnvelopedGenerator.RC2Cbc) {
-				var param = (DerInteger) identifier.Parameters;
-				int bits = param.Value.IntValue;
+                // RC2CBCParameter see: https://tools.ietf.org/html/rfc3370#section-5.2
+                var param = (DerSequence)identifier.Parameters;
+
+                var rc2ParameterVersion = (DerInteger)param[0];
+				int bits = rc2ParameterVersion.Value.IntValue;
 
 				switch (bits) {
-				case 128: algorithm = EncryptionAlgorithm.RC2128; return true;
-				case 64: algorithm = EncryptionAlgorithm.RC264; return true;
-				case 40: algorithm = EncryptionAlgorithm.RC240; return true;
+				    case 160: algorithm = EncryptionAlgorithm.RC240; return true;
+				    case 120: algorithm = EncryptionAlgorithm.RC264; return true;
+				    case 58: algorithm = EncryptionAlgorithm.RC2128; return true;
 				}
-			}
+            }
 
-			algorithm = EncryptionAlgorithm.RC240;
+            algorithm = EncryptionAlgorithm.RC240;
 
 			return false;
 		}


jstedfast:
Thanks!

jstedfast:
It seems that this fix creates a cast exception here:

```csharp
var param = (DerSequence) identifier.Parameters;
```

identifier.Parameters is a DerInteger in my unit tests which is testing a message generated by Thunderbird.

Do you actually have any real-world messages where this is a DerSequence?

12454j:
I never volunteered for this I'm not getting paid for it why me

James Thornton

jstedfast:
I'm not getting paid for this either.

As for why you, why are you getting this message? You aren't the original submitter for this pull request.

I was asking the original author of this pull request, not you.

12454j:
I am just trying  to rekindle a old interest and make some money

alxbauer:
Yes i had some real-world messages. That's why i found the issue ;-)
I will check next week if i got a message for you.

If identifier.Parameter is a DerInteger in your message, then it must be wrong.
According to RFC 3370 identifier.Parameter must be a sequence:

```
5.2  RC2 CBC

   The RC2 algorithm is described in RFC 2268 [RC2].  The algorithm
   identifier for RC2 in CBC mode is:

      rc2-cbc OBJECT IDENTIFIER ::= { iso(1) member-body(2) us(840)
          rsadsi(113549) encryptionAlgorithm(3) 2 }

   The AlgorithmIdentifier parameters field MUST be present, and the
   parameters field MUST contain a RC2CBCParameter:

      RC2CBCParameter ::= SEQUENCE {
        rc2ParameterVersion INTEGER,
        iv OCTET STRING  }  -- exactly 8 octets

   The RC2 effective-key-bits (key size) greater than 32 and less than
   256 is encoded in the rc2ParameterVersion.  For the effective-key-
   bits of 40, 64, and 128, the rc2ParameterVersion values are 160, 120,
   and 58 respectively.  These values are not simply the RC2 key length.
   Note that the value 160 must be encoded as two octets (00 A0), since
   the one octet (A0) encoding represents a negative number.
```

jstedfast:
Yea, I read that in the spec when you originally submitted the pull request which was one of the major reasons I figured the patch was correct :)

I was just wondering if maybe for whatever reason real-world usage (such as Thunderbird) were consistently using DerInteger instead.

I committed a fix yesterday that checks if the parameters are a DerInteger vs a DerSequence and handles them based on that. I gues that's how I'll have to handle this.

Anyway, knowing that you have messages that use DerSequence is nice to know. If you can get me a sample that I can use in the unit tests, that would be stellar :)

12454j:
I wish I could I don't have a computer to do this I am on my phone

alxbauer:
> I committed a fix yesterday that checks if the parameters are a DerInteger vs a DerSequence and handles them based on that. I gues that's how I'll have to handle this.

That's perfect! Today i found both cases in real-word messages!
Maybe i found a reason for this.
In RFC section 4.3.2 the RC2 Key Wrap is specified, and in this case the parameter is only a DerInteger.

However. I've seen the RC2-Algorithm only in SmimeCapabilities for now.
I don't think anyone these days will use this algorithm to encrypt a message.

```
4.3.2  RC2 Key Wrap

   A CMS implementation MAY support mixed key-encryption and content-
   encryption algorithms.  For example, a 128-bit RC2 content-encryption
   key MAY be wrapped with a 168-bit Triple-DES key-encryption key.
   Similarly, a 40-bit RC2 content-encryption key MAY be wrapped with a
   128-bit RC2 key-encryption key.

   RC2 key encryption has the algorithm identifier:

      id-alg-CMSRC2wrap OBJECT IDENTIFIER ::= { iso(1) member-body(2)
          us(840) rsadsi(113549) pkcs(1) pkcs-9(9) smime(16) alg(3) 7 }

   The AlgorithmIdentifier parameter field MUST be RC2wrapParameter:

      RC2wrapParameter ::= RC2ParameterVersion

      RC2ParameterVersion ::= INTEGER

   The RC2 effective-key-bits (key size) greater than 32 and less than
   256 is encoded in the RC2ParameterVersion.  For the effective-key-
   bits of 40, 64, and 128, the rc2ParameterVersion values are 160, 120,
   and 58 respectively.  These values are not simply the RC2 key length.
   Note that the value 160 must be encoded as two octets (00 A0),
   because the one octet (A0) encoding represents a negative number.

   RC2 128-bit keys MUST be used as key-encryption keys, and they MUST
   be used with the RC2ParameterVersion parameter set to 58.
```

jstedfast:
Aha! Thanks for this info :)

PR #360, opened 2017-11-25
Fixed bug: erroneously detected 'From' as Mbox marker

Terge3141:
I tried to parse my Mbox file generated by Thunderbird. At a certain point the "Formatexception: System.FormatException: Failed to parse message headers." was thrown. The problem wasn't the actual message but it turned out that the message before wasn't parsed correctly.
In this message the function MimeParser::IsMboxMarker erroneously returned true. The reason was that the line coincidently started with 'From ' but it belonged to the message (no mbox marker). I changed it and now it checks for 'From -'.
This doesn't fix the problem because theoretically a line could also start with 'From -'. However, it becomes less likely.

### DIFF
--- a/MimeKit/MimeParser.cs
+++ b/MimeKit/MimeParser.cs
@@ -576,7 +576,7 @@ namespace MimeKit {
 			if (allowMunged && *inptr == (byte) '>')
 				inptr++;
 
-			return *inptr++ == (byte) 'F' && *inptr++ == (byte) 'r' && *inptr++ == (byte) 'o' && *inptr++ == (byte) 'm' && *inptr == (byte) ' ';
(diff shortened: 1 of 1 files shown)


jstedfast:
This is no good. Thunderbird uses `From -` but many other mail clients do not use the `-`. Theoretically, mail clients are supposed to use `From <something that looks like an address> <something that looks like a date>`, but that's not how things work in practice.

`From ` lines are supposed to be munged in mbox files. In other words, they should be munged to start with `>From ` instead.

If Thunderbird isn't doing that, then that's a bug in Thunderbird.

Now what *might* be happening is that Thunderbird is using the Solaris mbox format which is slightly different from the standard BSD mbox format, where the client sets a `Content-Length` header to specify the number of octets (aka bytes) long the message is so that the parser can ignore `From ` lines until it has read X number of octets.

If that is the case, then you can enable this behavior in the `ParserOptions`. For more information, see here: http://www.mimekit.net/docs/html/P_MimeKit_ParserOptions_RespectContentLength.htm

Here's some interesting reading by the guy that wrote the original Netscape Mail app (which is now Thunderbird): https://www.jwz.org/doc/content-length.html

PR #368, opened 2018-01-17
Compiling with XBuild

rpc-scandinavia:
Compiling on Linux with XBuild generates a "Not implemented exception".
The use of a temp variable fixes that.

### DIFF
--- a/MimeKit/AsyncMimeParser.cs
+++ b/MimeKit/AsyncMimeParser.cs
@@ -293,7 +293,11 @@ namespace MimeKit {
 						while (*inptr != (byte) '\n')
 							inptr++;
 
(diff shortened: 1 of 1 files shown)


jstedfast:
Seems like a bug in Mono's C# compiler, mcs (not xbuild).

Which version of Mono are you using?

jstedfast:
Builds on Mono on Linux and Mac fine for me.

rpc-scandinavia:
Okay, I found the error.

My Debian had version 4.6 of Mono, which I think is the stock version.
I installed version 5.4.1.6 from "http://www.mono-project.com/download/#download-lin-debian", and now I can compile aswell.

Thank you.

jstedfast:
Awesome :)

PR #390, opened 2018-04-12
BodyBuilder, support for (optional) ContentTransferEncoding

304NotModified:
I like the BodyBuilder, but I think setting optionally the ContentTransferEncoding is important.

### DIFF
--- a/MimeKit/BodyBuilder.cs
+++ b/MimeKit/BodyBuilder.cs
@@ -108,6 +108,13 @@ namespace MimeKit {
 		public string HtmlBody {
 			get; set;
 		}
+		
+		/// <summary>
+        	/// Gets or sets the <see cref="ContentEncoding"/> for the <see cref="TextBody"/> and <see cref="HtmlBody"/> part (optional)
+        	/// </summary>
+        	public ContentEncoding? ContentTransferEncoding { 
+			get; set; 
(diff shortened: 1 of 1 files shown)


304NotModified:
I'm don't see why AppVeyor fails.

jstedfast:
AppVeyor sometimes fails when trying to upload packages to Coveralls. That can be ignored.

That said... I'm not convinced of this change. The way I advocate for setting the Content-Transfer-Encoding is via the use of `message.Prepare (EncodingConstraint.Value)`.

Granted, it seems that the BodyBuilder.Attachments.Add() logic sets a Content-Transfer-Encoding, but this is an artifact of a time before I had written `MimeMessage.Prepare()`.

If BodyBuilder starts setting the ContentTransferEncoding of the textual body parts, then it should use the same smarts as the Attachment logic and not leave it up to the client to "guess".

304NotModified:
> AppVeyor sometimes fails when trying to upload packages to Coveralls. That can be ignored.

@jstedfast  FYI, if that's a common issue with Coveralls, you could consider replacing  Coveralls with codecov.io (https://codecov.io/). I have codecov.io integration running without any problems for years.    codecov.io is also free for OSS.

I could help with the integration if desired .

PR #495, opened 2019-07-26
handle body consisting of only a single non-text entity as an attachment

naasking:
I recently switched my custom mail parser that I've been running for almost a decade over to MimeKit, and it's been running great, so thanks! I've found only a single problematic case in the email history I've checked, and it's where the body is empty except for a non-text entity, but whose content disposition header does not designate it as an attachment.

The change I made above straightforwardly only detects this one corner case. Let me know if you'd like to handle it a different way and I'll adapt it. I can provide a sample of this case if you need more clarification.

### DIFF
--- a/MimeKit/MimeMessage.cs
+++ b/MimeKit/MimeMessage.cs
@@ -976,9 +976,15 @@ namespace MimeKit {
 		/// <code language="c#" source="Examples\AttachmentExamples.cs" region="SaveAttachments" />
 		/// </example>
 		/// <value>The attachments.</value>
-		public IEnumerable<MimeEntity> Attachments {
-			get { return EnumerateMimeParts (Body).Where (x => x.IsAttachment); }
-		}
+		public IEnumerable<MimeEntity> Attachments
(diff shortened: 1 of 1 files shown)


jstedfast:
Here's the problem: everyone has a different understanding of what constitutes an attachment. Some people consider all images to be attachments, even if they are referenced by the HTML body, others do not because they consider them to be part of the body. Some consider all postscript content to be attachments, but what if the mail client wants to use postscript instead of HTML for rendering WYSIWYG content? Or an image. Or... some other type of content.

That's why MimeKit only considers MIME parts that are explicitly defined as "attachment" an attachment.

This is the same definition that the MIME specifications use and I think that's the only objective standard for deciding what is or is not an attachment that MimeKit itself can implement.

I would recommend taking a look at the HtmlPreviewVisitor class in the [FAQ](https://github.com/jstedfast/MimeKit/blob/master/FAQ.md#HasAttachments) which shows a better way to determine what MIME parts are attachments or not.

PR #509, opened 2019-09-23
Instantiate CmsSigner outside loop

### LINKED ISSUE #508: GetCmsSigner() opens multiple data readers
Hello,

the fix for issue #500 causes a problem with certain certificates databases.

`GetCmsSigner()` attempts to open a second data reader inside `BuildCertificateChain()`, while one is already active by `dbase.Find()`, this is not supported by many databases.

The easiest fix would be to instantiate the `CmsSigner` outside the loop.

A workaround for MsSQL certificate databases is to enable MultipleActiveResultSets, but no equvalient is available for MySQL.

swoga:
Closes #508


jstedfast:
Ah, committed my fix before I noticed you had sent in a PR. Still waking up out of my morning fog, I guess.

Patch looks good, though! :-)

PR #513, opened 2019-10-25
Fixed X509Crl.IsDelta for CRLs without extensions

swoga:
Hi,

when using revocation checking, the import for certain CRLs fails, because `GetCriticalExtensionOids()` inside `BouncyCastleCertificateExtensions.IsDelta` returns null for CRLs without any extensions.

This PR should fix this issue.

example: [crl without extensions](http://esignature.telekom.at/taca/crl/serialNumber_16_eSignature_Basic.crl)


jstedfast:
MimeKit 2.4.0 has just been released with this fix.

PR #551, opened 2020-04-02
Added explicit MIME type registation

naasking:
This change just provides an explicit registration method for a MIME type that may not be included in the default list.

I've been using my [Sasa.Mime](https://sourceforge.net/p/sasa/code/ci/default/tree/Sasa.Mime/) with this behaviour for years, and explicit registration has come in handy a few times when the MIME database I was working with was missing some Microsoft MIME types (like the one I added to the test).

Let me know if you're interested in any of the other features from Sasa.Mime. For instance, there's [a T4 template](https://sourceforge.net/p/sasa/code/ci/default/tree/Sasa.Mime/MediaTypePP.tt) which automatically downloads and creates the MIME database that's provided by default.


jstedfast:
I hope you don't mind, but I did some git surgery and manually merged your commits to be a bit cleaner: https://github.com/jstedfast/MimeKit/commit/a08a85a2bdce015310d88a921cf1da02388f9700

Looks good. Thanks!

PR #575, opened 2020-06-02
Update OpenPgpContext.cs

polterguy:
Howdy mate, I had to make the `GetSecretKey` method public and virtual. My reasons is that in order to lookup secret key(s) from a database (for instance, non GnuPG file system), it was necessary.

The reasons are that the default implementation of `GetSecretKey`, which is used when decrypting a `MultipartEncrypted`, it will invoke `EnumerateSecretKeys`, which obviously doesn't make much sense when you might have millions of keys. Hence, I realised if I made `GetSecretKey` **public and virtual**, I could use the `keyId` to lookup a specific key from my database, completely ignoring the entire idea of enumerating all keys.

Not sure if I am missing something here, but I think this is what's necessary in order to create my completely custom (database based) key storage ...

Hopefully you're ok with this :)

### DIFF
--- a/MimeKit/Cryptography/OpenPgpContext.cs
+++ b/MimeKit/Cryptography/OpenPgpContext.cs
@@ -884,7 +884,13 @@ namespace MimeKit.Cryptography {
 			throw new UnauthorizedAccessException ();
 		}
 
-		PgpSecretKey GetSecretKey (long keyId)
+
+		/// <summary>
+		/// Returns the secret key with the specified key ID
+		/// </summary>
+		/// <param name="keyId">ID of secret key to retrieve.</param>
+		/// <returns></returns>
+		public virtual PgpSecretKey GetSecretKey (long keyId)
 		{
 			foreach (var key in EnumerateSecretKeys ()) {
 				if (key.KeyId == keyId)


polterguy:
Psst, I figure as I encrypt MIME messages, I already know the recipient(s), and I can use `MultipartEncrypted.Encrypt` passing in an explicit array of `PgpSecretKey`s. And as sign a MIME message, I also know which key I want to sign with, so I can pass in my `PgpSecretKey`, explicitly using the override of `MultipartSigned.Create`that takes a PgpSecretKey.

However, the only place I actually don't know the key for sure, is during parsing of a MIME message, since I might have multiple private keys, due to some having been revoked, etc, at which point I'll need to lookup my own secret key for decrypting dynamically some way ...

Not sure if I missed something here, since I still haven't started looking at verifying signed messages yet (completely).

jstedfast:
Sure, this looks fine.

polterguy:
Great :)
When can I expect a NuGet ...? ;)
I desperately do *not* want to use the project directly in my own stuff. Prefer NuGet reference ...

Also, validating signatures in signed emails can be done using the `EnumeratePublicKey` override taking a `MailboxAddress`, right ...?
As in, I don't need to enumerate _all_ public keys I've got ...?
Sorry, just checking, to make sure I'll not need another patch to build my thing :)

jstedfast:
Yes, you can just override the one that takes a MailboxAddress.

PR #576, opened 2020-06-03
Making sure also public keys kan be retrieved without relying upon PublicKeyRingBundle

polterguy:
This too method needs to be public and virtual, in order for me to retrieve public keys, without relying upon `PublicKeyRingBundle`, which makes no sense, if user wants to store (millions potentially) of keys in database.

### DIFF
--- a/MimeKit/Cryptography/CryptographyContext.cs
+++ b/MimeKit/Cryptography/CryptographyContext.cs
@@ -44,7 +44,7 @@ namespace MimeKit.Cryptography {
 	{
 		const string SubclassAndRegisterFormat = "You need to subclass {0} and then register it with MimeKit.Cryptography.CryptographyContext.Register().";
 		static Func<SecureMimeContext> SecureMimeContextFactory;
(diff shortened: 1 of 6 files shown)


polterguy:
Jo mate, I took the liberty of doing what I proposed, to create an interception class between `OpenPgpContext` and `CryptographyContext`, to make it simpler to implement completely custom key persistence layer. Hopefully it's in accordance to your standards :)

polterguy:
BTW, not sure why the `TestAutoKeyRetrieve` test is failing, and I have firewall issues where I'm at, and I can't test it myself ... :/

polterguy:
Darn it! Thank you for CI ...!! :/

On its way, finally I think ... :)

jstedfast:
The TestAutoKeyRetrieve() test fails intermittently due to the fact that Werner's GnuPG key, depending on which keyserver is in the rotation, is sometimes signed with an Elliptic Curve key which is not yet supported by BouncyCastle.

https://github.com/bcgit/bc-csharp/issues/73

I would not worry about it failing.

Issue #591, opened 2020-07-15
Expose a way to tell why Arc Verification fails


The-Nutty:
**Is your feature request related to a problem? Please describe.**
It would be very good to be able to tell why Arc verification fails as currently there is no way to differentiate what failed when attempting to validate arc.

**Describe the solution you'd like**
`ArcVerifier.GetArcHeaderSets` https://github.com/jstedfast/MimeKit/blob/778ca3f1ac1d3edb16f9adb8894097218d37f742/MimeKit/Cryptography/ArcVerifier.cs#L546  already takes a `bool throwOnError` parameter that always false, this could be exposed in a Verify/VerifyAsync overload You then could make a few changes to VerifyAsync to throw if Signature or Seal validation fail or not, im my use case its really if GetArcHeaderSets throws that i care about. This should be quite simple and would not break existing functionality.

**Describe alternatives you've considered**
Could expand ArcSignatureValidationResult to contain other fail states but im not sure thats appropriate.

The-Nutty:
Furthure more another thing that would make the arc Verifier a big more usable is being able to verify signatures/seals even if cv=fail somewhere in the chain. 

Currently if cv=fail https://github.com/jstedfast/MimeKit/blob/778ca3f1ac1d3edb16f9adb8894097218d37f742/MimeKit/Cryptography/ArcVerifier.cs#L529 then mimekit does not bother to verify the seals, when infact for our use case that would be very useful.

Im happy to raise a new feature request for this if you would like but seemed logical here.

jstedfast:
Can you explain why you need this? Maybe if I had more information about what you are trying to do, I could come up with a better API?

I really don't want to expose a virtual method that takes a `throwOnError` argument. That would be a HUGE design failure in my opinion.

The typical use-case is "Does this message have a valid ARC signature chain? yes/no"

The-Nutty:
Yeah i understand that, for me it would be useful to know why arc failed. Some of this is routed in https://github.com/jstedfast/MimeKit/issues/590 as if ARC is failing we need to know why so it can be fixed. 

Our other use case is in an arc chain compromising of 3 intermediary. Where i=1 cv=none, i=2 cv=fail and i=3 cv=pass. It would be good to know if the signature/seal of i=3 passed even though for i=2 cv=fail.

jstedfast:
I guess what I'm asking is: do you want this for diagnostic purposes (i.e. to debug issues)? Or do you want it for some other reason?

[...]

PR #593, opened 2020-07-25
Added improved error reporting for ArcVerifier

jstedfast:
Fixes issue #591


jstedfast:
That's always been the case, but I get what you are saying. I'll change that if/when someone needs it. I believe the spec says not to sign when that's the case, but it's been a while since I looked, so I could be wrong. Not sure why I would have added that logic otherwise.

PR #610, opened 2020-10-07
Update to latest BouncyCastle and remove unsupported platforms (NS 1.x)

clairernovotny:
Removes Netstandard 1.x (all targets of which are no longer supported), and updates to the latest BouncyCastle version


jstedfast:
Yea, it's probably time to drop support for netstandard1.x... thanks!

PR #614, opened 2020-10-21
Include message-id

OzBob:
RFC 2822 as "optional" (although it says that it should be present):

Though optional, every message SHOULD have a "Message-ID:" field.

As per my bitter experience with my ISP, and just finding this post in SO: https://stackoverflow.com/questions/57837395/dkim-fails-when-sending-mails-with-smtplib

Adding a Message-Id ensured that my DKIM signature was valid and all my emails did not end up in Junk.

If this PR is accepted I'll also update some other samples with message-Id.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,7 @@ var message = new MimeMessage ();
 message.From.Add (new MailboxAddress ("Joey", "joey@friends.com"));
 message.To.Add (new MailboxAddress ("Alice", "alice@wonderland.com"));
 message.Subject = "How you doin?";
+message.MessageId = MimeKit.Utils.MimeUtils.GenerateMessageId();
 
 message.Body = new TextPart ("plain") {
     Text = @"Hey Alice,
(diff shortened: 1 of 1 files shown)


jstedfast:
This shouldn't be necessary, because MimeMessage already sets a Message-Id by default.

OzBob:
I agree, my mistake.

My code is:
```
var msg = MimeKit.MimeMessage.CreateFromMailMessage(mailmsg);
```
e.g. https://github.com/OzBob/SMTPMailClient/blob/master/SMTPMailClient/InsolMailKit.cs#L46

Which creates a msg without a MessageId

The source mailmsg object is of type ```System.Net.Mail.MailMessage``` which does not have a MessageId.

Would it be OK to add
```
if (!string.IsNullOrEmpty(msg.MessageId))
    msg.MessageId = MimeKit.Utils.MimeUtils.GenerateMessageId();
```
 to https://github.com/jstedfast/MimeKit/blob/f09c83906738727638d33a5fcd236531281e9ce6/MimeKit/MimeMessage.cs#L3171
?

jstedfast:
Is there a reason that you are constructing the message as a MailMessage and then converting it to a MimeMessage?

OzBob:
Haha, yes, i wish i wasn't but it is a shared library from an older asp.net website that used the previous .Net SMTP library and MailMessage.

jstedfast:
Ah, okay.

PR #616, opened 2020-11-03
Don't allocate when adding attachments.

JimBobSquarePants:
Hi, I don't know if you're looking for this kind of PR but I spotted the allocation when F12ing through the methods in Visual Studio.

A quick search for `new byte[` indicates that there are quite a few places in the library where the same pattern could be used e.g. `FilteredStream.Read`. Unfortunately I cannot currently dedicate the time to replace each instance so I thought I would introduce an example fix for this instance.

### DIFF
--- a/MimeKit/AttachmentCollection.cs
+++ b/MimeKit/AttachmentCollection.cs
@@ -136,18 +136,22 @@ namespace MimeKit {
 			if (attachment.ContentType.IsMimeType ("text", "*")) {
 				var filter = new BestEncodingFilter ();
 				const int BufLength = 4096;
-				var buf = ArrayPool<byte>.Shared.Rent (bufLength);
+				var buf  = ArrayPool<byte>.Shared.Rent (BufLength);
 				int index, length;
 				int nread;
 
-				while ((nread = stream.Read (buf, 0, bufLength)) > 0) {
-					filter.Filter (buf, 0, nread, out index, out length);
-					content.Write (buf, 0, nread);
+				try {
+
+					while ((nread = stream.Read (buf, 0, BufLength)) > 0) {
+						filter.Filter (buf, 0, nread, out index, out length);
+						content.Write (buf, 0, nread);
+					}
+
+					filter.Flush (buf, 0, 0, out index, out length);
+
+				} finally {
+					ArrayPool<byte>.Shared.Return (buf);
 				}
-
-				filter.Flush (buf, 0, 0, out index, out length);
-
-				ArrayPool<byte>.Shared.Return (buf);
 
 				attachment.ContentTransferEncoding = filter.GetBestEncoding (EncodingConstraint.SevenBit);
 			} else {
--- a/MimeKit/AttachmentCollection.cs
+++ b/MimeKit/AttachmentCollection.cs
@@ -31,6 +31,7 @@ using System.Collections.Generic;
 
 using MimeKit.IO;
 using MimeKit.IO.Filters;
+using System.Buffers;
 
 namespace MimeKit {
 	/// <summary>
@@ -134,16 +135,19 @@ namespace MimeKit {
 
 			if (attachment.ContentType.IsMimeType ("text", "*")) {
 				var filter = new BestEncodingFilter ();
-				var buf = new byte[4096];
+				int bufLength = 4096;
+				var buf = ArrayPool<byte>.Shared.Rent (bufLength);
 				int index, length;
 				int nread;
 
-				while ((nread = stream.Read (buf, 0, buf.Length)) > 0) {
+				while ((nread = stream.Read (buf, 0, bufLength)) > 0) {
 					filter.Filter (buf, 0, nread, out index, out length);
 					content.Write (buf, 0, nread);
 				}
 
 				filter.Flush (buf, 0, 0, out index, out length);
+
+				ArrayPool<byte>.Shared.Return (buf);
 
 				attachment.ContentTransferEncoding = filter.GetBestEncoding (EncodingConstraint.SevenBit);
 			} else {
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
[Failing test `TestReferenceByContentId`](https://dev.azure.com/jestedfa/MimeKit/_build/results?buildId=884&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=7dae026f-9e99-5d65-075e-3f7579577f94&l=42) does not seem to be related the change and fails with current master on my machine. 

``` bash
Message: 
      Contains failed.
      Expected: True
      But was:  False
    
  Stack Trace: 
    MultipartRelatedTests.TestReferenceByContentId() line 151
```

Indecently `TestDocumentRoot` fails on my machine also with the message.

``` bash
  Message: 
      Initial Root
      Expected: <X-MimeKit-Warning: Do NOT use ToString() to serialize entities! Use one of the WriteTo() methods instead!
    Content-Type: text/html; charset=utf-8
    Content-Id: <7F64657G3CU4.1KUI3FL529CE3@MjÃ¶lnir>
    
    This is the html body...>
      But was:  <X-MimeKit-Warning: Do NOT use ToString() to serialize entities! Use one of the WriteTo() methods instead!
    Content-Type: image/gif
    Content-Disposition: inline; filename=empty.gif
    Content-Id: <Z3H2657G3CU4.QIGWC03L02CU3@MjÃ¶lnir>
    
    >
    
  Stack Trace: 
    MultipartRelatedTests.TestDocumentRoot() line 70
```

MimeKit/AttachmentCollection.cs
```diff
@@ -134,17 +135,20 @@ static void LoadContent (MimePart attachment, Stream stream)
...
 
-				while ((nread = stream.Read (buf, 0, buf.Length)) > 0) {
+				while ((nread = stream.Read (buf, 0, bufLength)) > 0) {
 					filter.Filter (buf, 0, nread, out index, out length);
 					content.Write (buf, 0, nread);
 				}
 
 				filter.Flush (buf, 0, 0, out index, out length);
 
+				ArrayPool<byte>.Shared.Return (buf);
```
stevejgordon, line 150: I'd usually recommend wrapping the above code (while loop and flush) inside a try block and then putting the return inside a finally to ensure that it's returned despite potential exceptions. It'll get collected eventually but it's the pattern I've seen recommended.

JimBobSquarePants: If you think about it, if there's an unhandled exception then it doesn't matter whether the buffer is returned or not since the application will break. It's all managed memory so the GC will collect the buffer eventually.

I could add the pattern though if that's what everyone wants.

stevejgordon: I must admit, I've not explored the usage of this and whether exceptions may be caught higher up. I'd assumed that might be the case and if so, there's a slight benefit to putting it back in the pool cleanly.

MimeKit/AttachmentCollection.cs
```diff
@@ -134,17 +135,20 @@ static void LoadContent (MimePart attachment, Stream stream)
 
 			if (attachment.ContentType.IsMimeType ("text", "*")) {
 				var filter = new BestEncodingFilter ();
-				var buf = new byte[4096];
+				int bufLength = 4096;
```
antonfirsov, line 138:
```suggestion
				const int BufLength = 4096;
```

MimeKit/AttachmentCollection.cs
JimBobSquarePants: I'll add the pattern then (I haven't explored either tbh.)

jstedfast:
I've fixed that unit test.

Does `ArrayPool<byte>.Shared.Rent (BufLength);` sometimes return buffers of a length > than the size requested? I was wondering why use this constant if not.

jstedfast:
FWIW, I originally did not use System.Buffers because it was not available on all of the platforms that MimeKit originally supported.

I *think* that all of those .NET platforms have since been dropped, but I will need to verify.

If System.Buffers is available on all supported platforms, then I will be able to drop https://github.com/jstedfast/MimeKit/blob/master/MimeKit/Utils/BufferPool.cs (effectively the same idea).

JimBobSquarePants:
>Does ArrayPool<byte>.Shared.Rent (BufLength); sometimes return buffers of a length > than the size requested? I was wondering why use this constant if not.

@jstedfast Yes, the length can be greater so when working with buffers directly so you should ensure that you do not rely on that property. Ideally you'd wrap the pool and work directly with span slices only falling back to the array when you need it.

jstedfast (approved):

jstedfast:
I'll work on updating more areas of the code to use ArrayPool.

JimBobSquarePants:
Might be an idea to define some sort of implementation similar to [`IMemoryOwner<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.imemoryowner-1) to simplify pooling and length handling.

jstedfast:
Do you have an example of IMemoryOwner being used? Is the idea just to replace the need for the try/finally syntax and replacing it with a using block?

jstedfast:
BTW, just spotted this: `Content-Id: <Z3H2657G3CU4.QIGWC03L02CU3@MjÃ¶lnir>`

Is that your machine's hostname? Looks like I might need to encode the hostname if so. I sure hope it's that and not some sort of memory corruption.

JimBobSquarePants:
Yeah Mjölnir 😄

jstedfast:
Cool, cool :)

I just fixed GenerateMessageId() to encode the hostname if it is international.

PR #618, opened 2020-11-18
Added SQL Server Support

rblackin:
I needed to use this with SQL Server as the back end. I made changes to support it so I thought I would share the changes.

### DIFF
--- a/MimeKit/Cryptography/SqlCertificateDatabase.cs
+++ b/MimeKit/Cryptography/SqlCertificateDatabase.cs
@@ -229,32 +229,12 @@ namespace MimeKit.Cryptography {
 		/// <param name="column">The column to add.</param>
 		protected abstract void AddTableColumn (DbConnection connection, DataTable table, DataColumn column);
 
-		static string GetIndexName (string tableName, string[] columnNames)
+		public static string GetIndexName (string tableName, string[] columnNames)
 		{
 			return string.Format ("{0}_{1}_INDEX", tableName, string.Join ("_", columnNames));
 		}
-
-		static void CreateIndex (DbConnection connection, string tableName, string[] columnNames)
-		{
-			var indexName = GetIndexName (tableName, columnNames);
-			var query = string.Format ("CREATE INDEX IF NOT EXISTS {0} ON {1}({2})", indexName, tableName, string.Join (", ", columnNames));
-
-			using (var command = connection.CreateCommand ()) {
-				command.CommandText = query;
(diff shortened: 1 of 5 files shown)


MimeKit/Cryptography/SqlCertificateDatabase.cs
```diff
@@ -229,32 +229,12 @@ static DataTable CreateCrlsDataTable (string tableName)
 		/// <param name="column">The column to add.</param>
 		protected abstract void AddTableColumn (DbConnection connection, DataTable table, DataColumn column);
 
-		static string GetIndexName (string tableName, string[] columnNames)
+		public static string GetIndexName (string tableName, string[] columnNames)
 		{
```
jstedfast, line 233: Does this need to be `public` or can it be `protected`?

MimeKit/Cryptography/SqlCertificateDatabase.cs
```diff
@@ -229,32 +229,12 @@ static DataTable CreateCrlsDataTable (string tableName)
...
-			var indexName = GetIndexName (tableName, columnNames);
-			var query = string.Format ("DROP INDEX IF EXISTS {0}", indexName);
-
-			using (var command = connection.CreateCommand ()) {
-				command.CommandText = query;
-				command.ExecuteNonQuery ();
-			}
-		}
+		protected abstract void CreateIndex (DbConnection connection, string tableName, string[] columnNames);
+		protected abstract void RemoveIndex (DbConnection connection, string tableName, string[] columnNames);
```
jstedfast, line 238: It looks like it would have been a simpler change if you just made CreateIndex() and RemoveIndex() `protected virtual` instead of `protected abstract`. This would reduce code duplication between `SqliteCertificateDatabase` and `NpgsqlCertificateDatabase`.

jstedfast (changes requested):
Overall looks good and I'm so glad to see that it wasn't difficult for you to add! Excellent work!

MimeKit/Cryptography/SqlCertificateDatabase.cs
rblackin: switched from public. no reason for it to be static afaik.

MimeKit/Cryptography/SqlCertificateDatabase.cs
rblackin: good catch, changed

rblackin:
found more issues with sql server when adding a cert. Will recreate when fixed

PR #619, opened 2020-11-19
Implement Sql server support

rblackin:


jstedfast (approved):
Looks good! Thanks!

jstedfast:
Looks like the build failures were because the pipeline was using NuGet 5.3.0. When I bumped it to 5.8.0, it started working again. 🤷‍♂️ 

Looks like my X509CertificateDatabase API should have passed a DbConnection to all of the GetSelectCommand() and GetInsertCommand() methods.

I'll have to remedy that for v3.0 (I probably shouldn't break API for a minor release).

This is good, though, because w/o your patch, I might not have noticed!

PR #752, opened 2022-02-04
Improve Code Quality

iamcarbon:
- enables C# 10
- simplifies test project's csconfig.
- simplifies equality check on  TnefNameId
- removes obsolete pramas (and logic) for .NETSTANDARD1.x & NETCOREAPP30
- inlines variables declarations
- takes dependency on System.Memory to enable Span and eliminate various substring allocations


MimeKit/Cryptography/RsaEncryptionPadding.cs
```diff
@@ -243,25 +239,5 @@ internal AlgorithmIdentifier GetAlgorithmIdentifier ()
...
-				case DigestAlgorithm.Sha512: return RSAEncryptionPadding.OaepSHA512;
-				default: return null;
-				}
-			case RsaEncryptionPaddingScheme.Pkcs1:
-				return RSAEncryptionPadding.Pkcs1;
-			default:
-				return null;
-			}
-		}
-#endif
```
jstedfast, line 265: I might want to keep this block, but it should probably be `#if false` for now instead. Let me get back to you on this one... need to check the WindowsSecureMimeContext logic to see if this would be useful when converting MimeKit's CmsRecipient into a System.Security CmsRecipient. I *may* have a bug there if I don't have logic to do this kind of thing.

MimeKit/Cryptography/WindowsSecureMimeContext.cs
```diff
@@ -277,19 +277,10 @@ static RealCmsRecipientCollection GetCmsRecipients (CmsRecipientCollection recip
...
 					type = RealSubjectIdentifierType.SubjectKeyIdentifier;
 
-#if NETCOREAPP3_0
-				var padding = recipient.RsaEncryptionPadding?.AsRSAEncryptionPadding ();
-
-				if (padding != null)
-					real = new RealCmsRecipient (type, certificate, padding);
-				else
-					real = new RealCmsRecipient (type, certificate);
-#else
```
jstedfast, line 287: Ok, yea, these NETCOREAPP3_0 blocks need some closer examinations.

MimeKit/Text/HtmlAttributeId.cs
```diff
@@ -621,13 +621,9 @@ public static string ToAttributeName (this HtmlAttributeId value)
 		{
 			var name = value.ToString ();
 
-#if NETSTANDARD1_3 || NETSTANDARD1_6
-			var field = typeof (HtmlAttributeId).GetTypeInfo ().GetDeclaredField (name);
-			var attrs = field.GetCustomAttributes (typeof (HtmlAttributeNameAttribute), false).ToArray ();
-#else
+
```
jstedfast, line 624: I think this (green) line can be removed.

MimeKit/Utils/MimeUtils.cs
```diff
@@ -116,10 +115,7 @@ public static string GenerateMessageId (string domain)
 		/// <returns>The message identifier.</returns>
 		public static string GenerateMessageId ()
 		{
-#if NETSTANDARD1_3 || NETSTANDARD1_6
-			return GenerateMessageId ("localhost.localdomain");
-#else
-			if (DefaultHostName == null) {
+			if (DefaultHostName is null) {
 				try {
```
jstedfast, line 119: What are the implications of `is null` vs `== null`?

MimeKit/Cryptography/RsaEncryptionPadding.cs
```diff
@@ -182,8 +178,8 @@ public override string ToString ()
 		/// <returns><c>true</c> if <paramref name="left"/> and <paramref name="right"/> are equal; otherwise, <c>false</c>.</returns>
 		public static bool operator == (RsaEncryptionPadding left, RsaEncryptionPadding right)
 		{
-			if (ReferenceEquals (left, null))
-				return ReferenceEquals (right, null);
+			if (left is null)
+				return right is null;
```
jstedfast, line 183: I like the shorthand, but does `is` have any different overhead vs ReferenceEquals?

MimeKit/Cryptography/DefaultSecureMimeContext.cs
```diff
@@ -71,30 +71,22 @@ static DefaultSecureMimeContext ()
 		{
 			string path;
 
-#if !NETSTANDARD1_3 && !NETSTANDARD1_6
-			if (Path.DirectorySeparatorChar == '\\') {
+			if (Path.DirectorySeparatorChar is '\\') {
 				var appData = Environment.GetFolderPath (Environment.SpecialFolder.ApplicationData);
```
jstedfast, line 75: I'm more inclined to leave the `==`. Need to think about this.

MimeKit/MimeKit.csproj
```diff
@@ -6,6 +6,7 @@
     <VersionPrefix>3.1.1</VersionPrefix>
     <Authors>Jeffrey Stedfast</Authors>
     <TargetFrameworks>netstandard2.0;netstandard2.1;net452;net46;net47;net48;net5.0;net6.0</TargetFrameworks>
+    <LangVersion>10</LangVersion>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
jstedfast, line 10: I'm guessing this is needed for the `==` -> `is` changes?

MimeKit/MimeKit.csproj
```diff
@@ -51,14 +52,6 @@
...
 
-  <ItemGroup Condition=" $(TargetFramework.StartsWith('netstandard1.')) ">
-    <PackageReference Include="System.Reflection.TypeExtensions" Version="4.4.0" />
-    <PackageReference Include="System.Globalization.Extensions" Version="4.3.0" />
-    <PackageReference Include="System.Text.Encoding.CodePages" Version="4.4.0" />
-    <PackageReference Include="System.Data.Common" Version="4.3.0" />
-    <PackageReference Include="System.Buffers" Version="4.5.1" />
-  </ItemGroup>
-
   <ItemGroup Condition=" $(TargetFramework.StartsWith('netstandard2.')) ">
```
jstedfast, line 55: The MimeKitLite.csproj likely has a similar ItemGroup that can be removed.

UnitTests/UnitTests.csproj
```diff
@@ -43,308 +42,16 @@
...
-    <None Include="TestData\yenc\simple.msg" />
-    <None Include="TestData\yenc\state-changes.ntx" />
+    <MySQLiteInteropFiles Include="$(PkgStub_System_Data_SQLite_Core_NetFramework)\build\net46\**\SQLite.Interop.dll" />
   </ItemGroup>
 
   <ItemGroup>
-    <MySQLiteInteropFiles Include="$(PkgStub_System_Data_SQLite_Core_NetFramework)\build\net46\**\SQLite.Interop.dll" />
+    <None Remove="submit.ps1" />
+	<None Remove="test.ps1" />
   </ItemGroup>
```
jstedfast, line 55: These seem to have different indent levels.

jstedfast (changes requested):

MimeKit/Utils/MimeUtils.cs
iamcarbon: == null will use the type's == operator (when defined), while is null will check if the underlying object is null. Is null has a nice benefit of ensuring an equity operator doesn't accidentally change the behavior -- and can be generate better machine code in some cases.

MimeKit/Cryptography/RsaEncryptionPadding.cs
iamcarbon: Machine code is identical for these.

MimeKit/Cryptography/DefaultSecureMimeContext.cs
iamcarbon: the "is" operator nice for strings -- since it is an ordinal culture invariant comparison. Same as string.Equals("x", StringComparison.Ordinal);

For chars, codegen is the same -- but does unlock pattern matching against a variable name to enable:

c is 'a' or 'b'

If we utilize these patterns, we should consider utilizing is everywhere for consistency. I only changed a few right now to see how this settles with the project style. Let me know what you end up thinking.

MimeKit/MimeKit.csproj
iamcarbon: Yep -- better pattern matching, and better codegen on .NET6.0 for string interpolation. Our min targets hold us back from utilizing much else.

MimeKit/Cryptography/RsaEncryptionPadding.cs
iamcarbon: 👍

MimeKit/MimeKit.csproj
iamcarbon: looks like these already got removed there.

iamcarbon:
@jstedfast ready for review! thanks for the early start. the biggest change here is the introduction of System.Memory. I spend a few minutes eliminating a few allocations -- there's quite a few more we can eliminate if we take this dependency. 

The pattern matching changes are also to see how they feel in the project. If approved, it would be nice to make them consistent everywhere. We can also use the new "a switch" syntax, which would also be a nice cleanup.

jstedfast:
I'm going to cherry-pick your commits so I can manually squash them out of order (so fixups are squashed into their logical "parent")

MimeKit/Cryptography/RsaEncryptionPadding.cs
jstedfast: I reverted the NETCOREAPP3_0 stuff locally because it looks like I should really be using this stuff (but obviously not #if'd as NETCOREAPP3_0 - it'll probably be #if NET5_0 or something).

iamcarbon: glad we uncovered this (dead code that shouldn't have been killed)

iamcarbon:
@jstedfast Let me know what you decide on using is and taking a dependency on System.Memory. Happy to re-branch and re-submit those change as clean individual PRs.

jstedfast:
Thanks for helping to motivate me to finally get off my butt and start using spans. Been meaning to do this for years but was kinda stuck thanks to net45 for the longest time. Dropped support for it for 3.0, but forgot to address the span stuff I had wanted to do.

Also lets me finally get rid of my ICharArray hack :)

iamcarbon:
I think we got the good stuff from this PR merged! gonna close!

PR #753, opened 2022-02-05
Improve Code Quality 2

iamcarbon:
- Enables C# 10
- Uses is when comparing strings exactly (improving codegen)
- Eliminates two string allocations
- Uses switch expressions


iamcarbon:
@jstedfast Ready for review / feedback.

PR #754, opened 2022-02-05
Improve Code Quality 2

iamcarbon:
- Enables C# 10
-  Use is when comparing strings exactly (improving codegen)
- Eliminates two string allocations
- Uses switch expressions

### DIFF
--- a/MimeKit/ContentType.cs
+++ b/MimeKit/ContentType.cs
@@ -280,8 +280,8 @@ namespace MimeKit {
 			if (mediaSubtype == null)
 				throw new ArgumentNullException (nameof (mediaSubtype));
 
-			if (mediaType == "*" || mediaType.Equals (type, StringComparison.OrdinalIgnoreCase))
-				return mediaSubtype == "*" || mediaSubtype.Equals (subtype, StringComparison.OrdinalIgnoreCase);
+			if (mediaType is "*" || mediaType.Equals (type, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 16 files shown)


iamcarbon:
@jstedfast Isolated the syntax changes. Ready for review / feedback!

jstedfast:
I'm good with [1c33d42](https://github.com/jstedfast/MimeKit/pull/754/commits/1c33d423bedbdb228e5930209a93993c73038040).

I'm not liking the switch syntax changes (I prefer the old syntax, guess I'm just old school).

The `is` changes I'm still considering. My biggest hold-out right now is the dependency on C# 10 - I would prefer developers stuck on older VS would be able to build this. From a quick search, it *looks like* VS 2019, for example, only supports up to C# 8 (by default).

iamcarbon:
Notable: It looks like that we may have already broken Visual Studio 2019 by targeting .NET6.0. It looks like Microsoft bumped the SDK requirement to VS2022 at RTM. 

A few other source projects that have targeted .NET6.0 have had discussions on this recently. 
 
e.g. https://github.com/AvaloniaUI/Avalonia/issues/6955

With the ability to run multiple VS versions on the same machine, and a free Community edition used for open-source projects, the chances of being stuck are vastly reduced.

iamcarbon:
@jstedfast OK, new proposal is for us to use C# 8, introduced in the ValueStringBuilder PR! net452 is holding us back from using any of the good stuff anyhow. :)

Thanks for bearing through getting acclimated with the codebase and caught up its conventions. 

Closing this PR.

jstedfast:
Thanks for the net6 info, I did not think of that.

PR #755, opened 2022-02-06
Reduce Allocations

iamcarbon:
- Introduces and uses a StringBuilder.Append(StringBuilder) polyfill to eliminate an allocation
- Introduces a custom splitter to eliminate a few array allocations (and suballocations)
- Refactors the header tokenizer to eliminate allocations (looks like we both jumped on this -- and I need to rebase)

### DIFF
--- a/MimeKit/Cryptography/X509CertificateDatabase.cs
+++ b/MimeKit/Cryptography/X509CertificateDatabase.cs
@@ -40,6 +40,7 @@ using Org.BouncyCastle.Asn1.BC;
 using Org.BouncyCastle.Asn1.Pkcs;
 using Org.BouncyCastle.Asn1.X509;
 using Org.BouncyCastle.X509.Store;
+using MimeKit.Utils;
 
 namespace MimeKit.Cryptography {
 	/// <summary>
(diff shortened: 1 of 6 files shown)


iamcarbon:
@jstedfast We should consider replacing Word with ReadOnlyMemory<char> (same structure / shape)

iamcarbon:
I'll rebase this on-top of your work a bit later. Notable nits are the StringBuilder.Append(StringBuilder) polyfill and custom splitting logic. 

I've done a couple reviews of the library at this point -- the code is pretty darn optimized! Looks like were chomping at low level allocations at this point.  :)

iamcarbon:
Next big possible optimization would be to use a ValueStringBuilder to eliminate all these StringBuilder allocations.

jstedfast:
One thing we used in Visual Studio for Mac when I was on that team (I'm on VS Windows now) was a StringBuilder cache, which might be another possibility.

jstedfast:
I like the Splitter stuff. Maybe call it StringSplitter?

This was all great stuff!

jstedfast:
As you noted, I rewrote the Header TokenizeText method after being inspired by your changes. I had to get in on the Span action 😁

We can move the Break() method back out to be a static method like before if we can switch over to using a built-in struct/class. We don't have to use my Word struct.

I think I did that because I couldn't use IEnumerable<ReadOnlySpan<char>> as a return type.

PR #756, opened 2022-02-06
Introduce ValueStringBuilder

iamcarbon:
This PR introduces and uses a ValueStringBuilder via [dotnet](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/Common/src/System/Text/ValueStringBuilder.cs).

Under the hood, this uses a dynamically resizing array borrowed from ArrayPool to eliminate allocations. It can also accept stack allocated memory as an initial buffer (expanding to use borrowed heap memory, as needed). 

We must be careful to return the memory after use by calling ToString or Dispose, but otherwise -- this should be a nice net win for throughput.

### DIFF
--- a/MimeKit/Header.cs
+++ b/MimeKit/Header.cs
@@ -650,7 +650,7 @@ namespace MimeKit {
 		{
 			var tokens = new List<ReceivedTokenValue> ();
 			var rawValue = encoding.GetBytes (value);
-			var encoded = new StringBuilder ();
+			var encoded = new ValueStringBuilder (128);
 			int lineLength = field.Length + 1;
 			bool date = false;
 			int index = 0;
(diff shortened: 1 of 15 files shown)


iamcarbon:
@jstedfast Ready for review + feedback. I've also enabled C# 8 so we can maintain source compatibility with ValueStringBuilder and the official dotnet source to sync future changes / improvements.

MimeKit/Header.cs
```diff
@@ -531,12 +531,12 @@ public string GetValue (string charset)
...
 		{
-			var encoded = new StringBuilder (" ");
-			int lineLength = field.Length + 2;
-
 			if (!InternetAddressList.TryParse (options, value, out var list))
 				return (byte[]) format.NewLineBytes.Clone ();
 
+			var encoded = new StringBuilder (" ");
+			int lineLength = field.Length + 2;
+
```
jstedfast, line 539: 👍

jstedfast (approved):

PR #758, opened 2022-02-07
Reduce Allocations (take 2)

iamcarbon:
- Introduces and uses a StringSplitter to avoid array (and array element) allocations
- Use ValueStringBuilder.AppendInvariant(value) to eliminate allocations appending span formattables on NET6+
- Eliminates various byte[] and string allocations

### DIFF
--- a/MimeKit/Utils/MimeUtils.cs
+++ b/MimeKit/Utils/MimeUtils.cs
@@ -24,8 +24,6 @@
 // THE SOFTWARE.
 //
 
-#nullable enable
-
 using System;
 using System.Text;
 using System.Collections.Generic;
@@ -42,7 +40,7 @@ namespace MimeKit.Utils {
 	public static class MimeUtils
 	{
 		const string base36 = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
-		static string? DefaultHostName = null;
+		static string DefaultHostName = null;
 
 		/// <summary>
 		/// A string comparer that performs a case-insensitive ordinal string comparison.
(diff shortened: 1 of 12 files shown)


iamcarbon:
@jstedfast Ready for review!

MimeKit/Utils/MimeUtils.cs
```diff
@@ -86,10 +86,16 @@ public static string GenerateMessageId (string domain)
 
 			ulong value = (ulong) DateTime.UtcNow.Ticks;
 			var id = new ValueStringBuilder (64);
+
+#if NET6_0_OR_GREATER
+			Span<byte> block = stackalloc byte[8];
+
```
jstedfast, line 90: Can we stackalloc in both cases?

jstedfast (approved):

MimeKit/Utils/MimeUtils.cs
iamcarbon: sadly, no RandomNumberGenerator overload accepting spans on older runtimes.

jstedfast: No, I mean, instead calling `new byte[8]`, can we do `stackalloc byte[8]` in the non-6_0_OR_GREATER case?

iamcarbon: Looks like the subsequent call is hidden from the PR. byte[] block is passed into GetRandomBytes(byte[]). This method cannot be spanified, so we can't stackalloc block. 


```csharp
var block = new byte[8];

GetRandomBytes (block);
```

jstedfast:
This is all really great stuff, thanks!

PR #759, opened 2022-02-08
[WIP] Improve MimeParser safety

iamcarbon:
Work toward the spanification of MimeParser. Posting for early feedback.

### DIFF
--- a/MimeKit/MimeParser.cs
+++ b/MimeKit/MimeParser.cs
@@ -561,15 +561,6 @@ namespace MimeKit {
 			MimeEntityEnd?.Invoke (this, args);
 		}
 
-#if DEBUG_PARSER
-		static string ConvertToCString (byte[] buffer, int startIndex, int length)
-		{
-			var cstr = new StringBuilder ();
-			cstr.AppendCString (buffer, startIndex, length);
-			return cstr.ToString ();
-		}
-#endif
-
 		static int NextAllocSize (int need)
 		{
 			return (need + 63) & ~63;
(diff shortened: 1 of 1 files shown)


jstedfast:
I think this is all unnecessary, but I will inspect closer after work to verify.

I know it seems like ABR's are easily possible, but I'm fairly (90%?) certain they aren't.

jstedfast:
There are 5 uses of IsMboxMarker:

1. [Line #778](https://github.com/jstedfast/MimeKit/blob/3.1.1/MimeKit/MimeParser.cs#L778): This is safe because we check `markerLength >= 5` and allowMunged is `false`.
2. [Line #928](https://github.com/jstedfast/MimeKit/blob/3.1.1/MimeKit/MimeParser.cs#L928): We also check `length >= 5` and allowMunged is `false`.
3. [Line #939](https://github.com/jstedfast/MimeKit/blob/3.1.1/MimeKit/MimeParser.cs#L939): we check `length < 5` (if < 5 we don't call IsMboxMarker), but we do use allowMunged = `true` so we theoretically could compare up to 6 input bytes. I'll investigate this a bit more later, but I'm pretty sure even this one is safe (even if only because `MimeParser.input` has 4 bytes of unused padding at the end).
4. [Line #1194](https://github.com/jstedfast/MimeKit/blob/3.1.1/MimeKit/MimeParser.cs#L1194): We check `length >= 5` and allowMunged = `false`, so this isn't an issue.
5. [Line #1211](https://github.com/jstedfast/MimeKit/blob/3.1.1/MimeKit/MimeParser.cs#L1211): This is in the `IsBoundary()` method which only gets called when we've located the end of the line, so there will be a `'\n'` byte which will cause the `"From "` check to fail and prevent any possibility of an ABR (even discounting the fact that the input buffer has padding).

PR #764, opened 2022-02-13
Account for endianness when comparing uint

iamcarbon:
Ensures we account for endianness when reading uint. Ready for review.


iamcarbon:
We may be able to use this method for all "From " comparisons. Looks to be quite a few.

PR #773, opened 2022-02-16
Eliminate various string allocations

iamcarbon:
- Use CultureInfo.InvariantCulture when formatting integers
- Eliminates various string allocations

### DIFF
--- a/MimeKit/Cryptography/BouncyCastleCertificateExtensions.cs
+++ b/MimeKit/Cryptography/BouncyCastleCertificateExtensions.cs
@@ -195,7 +195,7 @@ namespace MimeKit.Cryptography {
 			var hex = new ValueStringBuilder (blob.Length * 2);
 
 			for (int i = 0; i < blob.Length; i++)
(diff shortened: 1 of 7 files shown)


iamcarbon:
@jstedfast ready for review.

MimeKit/Cryptography/BouncyCastleCertificateExtensions.cs
```diff
@@ -195,7 +195,7 @@ internal static string AsHex (this byte[] blob)
 			var hex = new ValueStringBuilder (blob.Length * 2);
 
 			for (int i = 0; i < blob.Length; i++)
-				hex.Append (blob[i].ToString ("x2"));
+				hex.AppendInvariant (blob[i], "x2");
```
iamcarbon, line 198: Note, we also fixed the format overload in ValueStringBuilder. This uses ISpanFormattable on .NET6.0+

jstedfast:
Thanks so much for all of these patches. Great stuff!

PR #794, opened 2022-05-19
Fix `MemoryBlockStream.Read` method.

fyodorklimenko:
`Math.Min` may return a negative value when the difference between
`length` and `position` is not in the range of `int`. It happens because
of the numeric conversion from `long` to `int`.

As a result, the `Read` method does not read a single byte.

This situation is possible when the length of the stream is greater than
`int.Max`.


jstedfast (approved):

jstedfast:
Awesome, thanks!

PR #796, opened 2022-05-29
Compression: improve memory usage. 

fyodorklimenko:
The current implementation of the compression algorithm gets encoded
data using the `CmsCompressedData.GetEncoded` method:
https://github.com/bcgit/bc-csharp/blob/release-1.9.0/crypto/src/cms/CMSCompressedData.cs#L105

This method delegates real work to the `Asn1Encodable` class which
initializes one more temporary `MemoryStream` to get an array of
bytes:
https://github.com/bcgit/bc-csharp/blob/release-1.9.0/crypto/src/asn1/Asn1Encodable.cs#L30

Suggested solution writes encoded data directly into the stream.


jstedfast (approved):
Awesome!

PR #806, opened 2022-06-17
Remove unused pragmas

iamcarbon:
Now that we've dropped net451, we can also remove some legacy logic.


jstedfast (approved):

jstedfast:
Thanks! I saw those the other day and meant to remove them as soon as I had free time again.

PR #807, opened 2022-06-22
Added two overloads that don't require a name

RobThree:
In many cases I don't care about an actual name; when some monitoring or logging system or a 'bot' is sending email there is no 'real name' anyway - setting the name equal to the address is fine (for me) in that situation. This commit adds two 'convenience constructors' which do away with the name argument and set the name to the given address by calling the overloaded constructors with the address value for the name argument.

### DIFF
--- a/MimeKit/MailboxAddress.cs
+++ b/MimeKit/MailboxAddress.cs
@@ -215,7 +215,7 @@ namespace MimeKit {
 		/// <exception cref="ParseException">
 		/// <paramref name="address"/> is malformed.
 		/// </exception>
-		public MailboxAddress (string address) : this (Encoding.UTF8, address, address)
+		public MailboxAddress (string address) : this (address, address)
 		{
 		}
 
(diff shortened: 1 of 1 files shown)


jstedfast:
I appreciate the PR, but I'm not sure this really makes sense to have in MimeKit.

It's easy enough to just use `string.Empty` as the name and having just a `string address` argument makes people think they can pass in strings like `User Name <user@example.com>` and have the .ctor parse it which is not what it does.

RobThree:
> I appreciate the PR, but I'm not sure this really makes sense to have in MimeKit.

No problem!

> It's easy enough to just use `string.Empty` as the name and having just a `string address` argument makes people think they can pass in strings like `User Name <user@example.com>` and have the .ctor parse it which is not what it does.

I think that would be an odd or weird assumption but it's ok to disagree. I'll use `string.Empty` instead. 👍

jstedfast:
I think it's because System.Net.Mail's MailAddress .ctor parses it whereas I have MailboxAddress.Parse() to handle that :-\

But yea, otherwise I also agree it's a weird assumption.

PR #881, opened 2023-01-26
[Tests] Enable ImplictUsings

iamcarbon:
This PR enables implicit usings on the test project.

NOTE: the bracing style is inconsistent within the test project. Most files in UnitTests.Cryptography have the brace on a new line -- a few updates were made to make the bracing style consistent within the file.

We may also consider updating the test project to use file-scoped namespaces to remove this inconsistency all together.

### DIFF
--- a/UnitTests/ArgumentExceptionTests.cs
+++ b/UnitTests/ArgumentExceptionTests.cs
@@ -24,19 +24,16 @@
 // THE SOFTWARE.
 //
 
-using System;
-using System.IO;
 using System.Reflection;
 
-using NUnit.Framework;
-
 using MimeKit;
-using MimeKit.IO;
-using MimeKit.Utils;
-using MimeKit.IO.Filters;
 using MimeKit.Cryptography;
+using MimeKit.IO;
+using MimeKit.IO.Filters;
+using MimeKit.Utils;
 
-namespace UnitTests {
+namespace UnitTests
+{
 	[TestFixture]
 	public class ArgumentExceptionTests
 	{
--- a/UnitTests/AssortedTests.cs
+++ b/UnitTests/AssortedTests.cs
@@ -24,18 +24,14 @@
 // THE SOFTWARE.
 //
 
-using System;
-using System.IO;
-using System.Linq;
-using System.Text;
 using System.Globalization;
-
-using NUnit.Framework;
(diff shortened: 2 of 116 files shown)


iamcarbon:
@jstedfast Ready for feedback. Also, let me know your thoughts on using file-scoped namespaces within the test project. File scoped namespaces make for a nice improvement in readability -- particularly on laptops. This will also eliminate some inconsistencies in bracing style there.

jstedfast:
I'm not a big fan of the newer file-scoped namespaces, but it's probably just because I'm not used to it.

I have similar feelings about the implicit usings :(

I feel like the using statements help me figure out where a particular type is coming from (or at least a handful of potential places).

For the namespace braces, I think I was trying to use the following style:

```csharp
namespace UnitTests.Cryptography {
```

My preferred bracing style is admitedly bizarre to most people. Braces on new lines for classes/structs/methods, but braces on the same line for everything else. It largely comes from my C coding style which is more-or-less based on the K&R style (or maybe more accurately 1TBS?), I think.

All of the code I write and/or deal with at Microsoft uses braces on a new line for literally everything and that seems to be the norm for everyone writing C#, at least (except me, haha).

I think the namespace bracing is inconsistent because I'm less opinionated on that one.

I'll keep this open for a bit longer to give myself time to think about implicit usings a bit more. I can sense that the trend will be to move toward ImplicitUsings=True, so maybe I should really think about it.

iamcarbon:
@jstedfast It's taken me a while to come around on ImplictUsings myself, and I'm also with you on being on where types are coming from in the main libraries. In my own projects, I usually limit the global using scopes to System & System.Collections.Generic. I'm particularly cautious in introducing a dependency on System.Linq (and accidentally capturing variables / and limiting trimming opportunities).

In tests, however -- I've actually found using the default ImplictUsing sets (which includes System, Linq, Threading, etc) -- to have a nice side effect of ensuring that the library never conflicts with a common System type name. It's also nice always having the Linq extensions in scope there, and having to scroll less when coding on a laptop.

Ponder on it!

jstedfast:
@iamcarbon Thanks for your insights - I will!

You make a good argument.

jstedfast:
I ended up merging this manually since it had conflicts with master

PR #884, opened 2023-02-13
Add a type converter for the InternetAddress type

0xced:
This is useful to automatically convert strings to typed InternetAddress subclasses. For example, when using the `Microsoft.Extensions.Configuration` package.

### DIFF
--- a/MimeKit/InternetAddress.cs
+++ b/MimeKit/InternetAddress.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // InternetAddress.cs
 //
 // Author: Jeffrey Stedfast <jestedfa@microsoft.com>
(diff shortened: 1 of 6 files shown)


jstedfast:
I'll have to think about this...

jstedfast:
Can the TypeConverter classes be internal or do they have to be public for this to work?

0xced:
I think they could be internal but we would lose the ability to specify another ParserOptions with the `Register` method.

Also, all the system type converter classes are public.

jstedfast:
Can you show me an example of how you intend to use this?

I'm not very familiar with Microsoft.Extensions.Configuration and so I'm curious about how this would work.

0xced:
Will do next week when I'm back in front of a computer.

jstedfast:
@0xced Just figured I'd poke you since you are likely back from vacation now.

jstedfast:
I made some changes to this PR and manually merged it (couldn't push my changes to this PR branch).

PR #888, opened 2023-02-20
Improve Code Quality

iamcarbon:
This PR: 

- Enables C# 10
- Prefers is null to == null (to ensure we don't call an overloaded operator)
- Remove unnecessary null checks on value structs
- Fixes a few typos
- Uses compound assignment, pattern matching, and inline variable decorations in a few places


iamcarbon:
@jstedfast Ready for review

jstedfast:
Thanks, these look good.

PR #894, opened 2023-02-28
RFC: Improve Code Quality

iamcarbon:
@jstedfast This PR utilizes a few language features to simplify the code and fixes a few misspellings.

Posting for feedback / comments. Let me know if these fit with the style of the library.

### DIFF
--- a/MimeKit/AttachmentCollection.cs
+++ b/MimeKit/AttachmentCollection.cs
@@ -86,9 +86,7 @@ namespace MimeKit {
 		/// Indicates the number of attachments in the collection.
 		/// </remarks>
 		/// <value>The number of attachments.</value>
-		public int Count {
-			get { return attachments.Count; }
-		}
+		public int Count => attachments.Count;
 
 		/// <summary>
 		/// Get whther or not the collection is read-only.
@@ -97,9 +95,7 @@ namespace MimeKit {
 		/// A <see cref="AttachmentCollection"/> is never read-only.
 		/// </remarks>
 		/// <value><c>true</c> if the collection is read only; otherwise, <c>false</c>.</value>
-		public bool IsReadOnly {
-			get { return false; }
-		}
+		public bool IsReadOnly => false;
 
 		/// <summary>
 		/// Get or set the <see cref="MimeEntity"/> at the specified index.
--- a/MimeKit/ContentType.cs
+++ b/MimeKit/ContentType.cs
@@ -231,9 +231,7 @@ namespace MimeKit {
 		/// Gets the simple mime-type.
 		/// </remarks>
 		/// <value>The mime-type.</value>
-		public string MimeType {
-			get { return string.Format ("{0}/{1}", MediaType, MediaSubtype); }
-		}
+		public string MimeType => $"{MediaType}/{MediaSubtype}";
 
 		/// <summary>
 		/// Get or set the name parameter.
--- a/MimeKit/Cryptography/AuthenticationResults.cs
+++ b/MimeKit/Cryptography/AuthenticationResults.cs
@@ -495,7 +495,7 @@ namespace MimeKit.Cryptography {
 
 				value = Encoding.ASCII.GetString (text, tokenIndex, index - tokenIndex);
 
-				if (value == "reason" || value == "action") {
+				if (value is "reason" or "action") {
 					if (!ParseUtils.SkipCommentsAndWhiteSpace (text, ref index, endIndex, throwOnError))
 						return false;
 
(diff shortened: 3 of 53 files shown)


MimeKit/AttachmentCollection.cs
```diff
@@ -86,9 +86,7 @@ public AttachmentCollection () : this (false)
 		/// Indicates the number of attachments in the collection.
 		/// </remarks>
 		/// <value>The number of attachments.</value>
-		public int Count {
-			get { return attachments.Count; }
-		}
+		public int Count => attachments.Count;
```
jstedfast, line 89: My OCD has so far prevented me from really liking the `=>` style of property declaration, but I'll think about it.

Over the past few years at work, I've been bombarded with this style all over the place (but it's generally inconsistent because "new" code will use this style but older code remains in the previous style, so we have a mish-mash of these styles).

MimeKit/ContentType.cs
```diff
@@ -231,9 +231,7 @@ public ContentType (string mediaType, string mediaSubtype)
 		/// Gets the simple mime-type.
 		/// </remarks>
 		/// <value>The mime-type.</value>
-		public string MimeType {
-			get { return string.Format ("{0}/{1}", MediaType, MediaSubtype); }
-		}
+		public string MimeType => $"{MediaType}/{MediaSubtype}";
```
jstedfast, line 235: I would definitely take the `$"{MediaType}/{MediaSubtype}"` over string.Format().

MimeKit/Encodings/Base64Decoder.cs
```diff
@@ -79,12 +79,11 @@ public Base64Decoder ()
...
-			decoder.saved = saved;
-			decoder.bytes = bytes;
-
-			return decoder;
+			return new Base64Decoder {
+				previous = previous,
+				saved = saved,
+				bytes = bytes
+			};
 		}
```
jstedfast, line 87: Stuff like this I will definitely take.

MimeKit/GroupAddress.cs
```diff
@@ -111,7 +111,7 @@ public GroupAddress (string name) : this (Encoding.UTF8, name)
 		/// <returns>The cloned group address.</returns>
 		public override InternetAddress Clone ()
 		{
-			return new GroupAddress (Encoding, Name, Members.Select (x => x.Clone ()));
+			return new GroupAddress (Encoding, Name, Members.Select (static x => x.Clone ()));
 		}
```
jstedfast, line 115: I've never seen static used that way. What does it mean in this context?

MimeKit/IO/ChainedStream.cs
```diff
@@ -172,15 +172,13 @@ void CheckCanWrite ()
...
-		}
+		public override bool CanTimeout => false;
 
 		/// <summary>
 		/// Get the length of the stream, in bytes.
 		/// </summary>
 		/// <remarks>
-		/// The length of a <see cref="ChainedStream"/> is the combined lenths of all
+		/// The length of a <see cref="ChainedStream"/> is the combined lengths of all
 		/// of its chained streams.
```
jstedfast, line 182: Thanks for the spelling fix!

MimeKit/GroupAddress.cs
iamcarbon: It ensures that the method doesn't capture any outside variables and creating a closure / allocation. It also creates a new scope for the variable names. In this case, we weren't capturing any outside variables, so the codegen is the same.

MimeKit/Cryptography/AuthenticationResults.cs
```diff
@@ -495,7 +495,7 @@ static bool TryParseMethods (byte[] text, ref int index, int endIndex, bool thro
 
 				value = Encoding.ASCII.GetString (text, tokenIndex, index - tokenIndex);
 
-				if (value == "reason" || value == "action") {
+				if (value is "reason" or "action") {
 					if (!ParseUtils.SkipCommentsAndWhiteSpace (text, ref index, endIndex, throwOnError))
```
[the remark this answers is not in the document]
iamcarbon: Yep -- the code gen is identical to ==  (Ordinal comparison).

MimeKit/AttachmentCollection.cs
iamcarbon: I tend to use expression bodies when there's a single line of logic, and use blocks for anything more complicated. But they do look out of place when mixed. I've omitted these changes in #895

iamcarbon:
Let me know what you think of the pattern matching changes too.

```
if (items[0] == 1 || items[0] == 2)
```

vs 

```
if (items[0] is 1 or 2)
```

MimeKit/GroupAddress.cs
jstedfast: Ah, gotcha. Seems like a good feature to keep in mind. Thanks.

PR #898, opened 2023-03-03
Simplify header licence

iamcarbon:
This PR updates the header license to use the minimal format used on .NET Foundation projects, and the same format used on other .NET projects (runtime, aspnetcore, ML.net, etc). This also eliminates the need to update the copyright year in the future.

Examples in other projects:

https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Http/src/DefaultHttpClientFactory.cs
https://github.com/dotnet/aspnetcore/blob/main/src/Azure/AzureAppServicesIntegration/src/AppServicesWebHostBuilderExtensions.cs

### DIFF
--- a/MimeKit/AddressParserFlags.cs
+++ b/MimeKit/AddressParserFlags.cs
@@ -1,28 +1,5 @@
-﻿//
-// AddressParserFlags.cs
-//
-// Author: Jeffrey Stedfast <jestedfa@microsoft.com>
-//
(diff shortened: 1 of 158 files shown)


iamcarbon:
@jstedfast Ready for feedback.

jstedfast:
Nah, I prefer the current license header.

PR #910, opened 2023-03-28
Signature expiration

Saalin:
It would be great if MimeKit supported signature expiration in DKIM signature header, as described in https://www.ietf.org/rfc/rfc6376.txt (page 24).

I will gladly apply any needed changes.

### DIFF
--- a/MimeKit/Cryptography/DkimSignerBase.cs
+++ b/MimeKit/Cryptography/DkimSignerBase.cs
@@ -135,6 +135,17 @@ namespace MimeKit.Cryptography {
 		public DkimCanonicalizationAlgorithm BodyCanonicalizationAlgorithm {
 			get; set;
 		}
+		
+		/// <summary>
+		/// Get or set the signature expiration offset.
+		/// </summary>
+		/// <remarks>
+		/// Get or set the signature expiration offset.
+		/// </remarks>
+		/// <value>The signature expiration offset value.</value>
+		public TimeSpan? SignatureExpirationOffset {
+			get; set;
+		}
 
 		/// <summary>
 		/// Get or set the canonicalization algorithm to use for the message headers.
--- a/MimeKit/Cryptography/DkimSignerBase.cs
+++ b/MimeKit/Cryptography/DkimSignerBase.cs
@@ -137,13 +137,13 @@ namespace MimeKit.Cryptography {
 		}
 		
 		/// <summary>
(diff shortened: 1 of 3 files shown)


MimeKit/Cryptography/DkimSignerBase.cs
```diff
@@ -135,6 +135,17 @@ protected DkimSignerBase (string domain, string selector, DkimSignatureAlgorithm
...
+		/// <summary>
+		/// Get or set the signature expiration offset.
+		/// </summary>
+		/// <remarks>
+		/// Get or set the signature expiration offset.
+		/// </remarks>
+		/// <value>The signature expiration offset value.</value>
+		public TimeSpan? SignatureExpirationOffset {
+			get; set;
+		}
```
jstedfast, line 148: Thanks for adding the documentation comment!

The specification defines the `x=` value as an absolute timestamp and I wonder if the API should do the same (although I think I understand your logic & reasoning for using an offset instead).

Even if we decide to stick with the TimeSpan API, I'm not sure I like the `XyzOffset` naming convention. I'm just thinking out loud here, but I wonder if `SignatureExpiration` or `SignatureExpiresAfter` might be better names?

I was just thinking about the X509Certificate2 API but `NotAfter` property name doesn't really work here.

I also wonder if we should pluralize `Signature` to `Signatures` since a signer could potentially be used to generate multiple signatures. e.g. `SignaturesExpireAfter`

Saalin: I think it would be more straightforward to simply copy RFC and use absolute timestamp, but isn't TimeSpan API more natural? Mentioning the example with multiple signatures, it would end with multiple messages with different `t=`, but the same `x=`, which seems to be weird.

I think both `Offset` nad `After` describes the intention, so it's fine for me to change. Also pluralizing seems to be good idea (I didn't think about reusing Signer).

jstedfast: > but isn't TimeSpan API more natural? Mentioning the example with multiple signatures, it would end with multiple messages with different t=, but the same x=, which seems to be weird.

Yes, I would tend to agree. I was just thinking out loud.

> I think both Offset nad After describes the intention, so it's fine for me to change. Also pluralizing seems to be good idea (I didn't think about reusing Signer).

👍

PR #917, opened 2023-05-15
Not parsing address headers in Looser RfcComplianceMode

grzegorz-pasternak:
Would it have sense to have such "feature"? We have some memory problems when processing TO header with 80k addressees. It is not important for us to do any analysis what's inside and parse all participants.

### DIFF
--- a/MimeKit/Header.cs
+++ b/MimeKit/Header.cs
@@ -557,7 +557,7 @@ namespace MimeKit {
 
 		static byte[] EncodeAddressHeader (ParserOptions options, FormatOptions format, Encoding encoding, string field, string value)
 		{
-			if (!InternetAddressList.TryParse (options, value, out var list))
+			if (options.AddressParserComplianceMode == RfcComplianceMode.Looser || !InternetAddressList.TryParse (options, value, out var list))
 				return EncodeUnstructuredHeader (options, format, encoding, field, value);
 
 			var encoded = new StringBuilder (" ");


jstedfast:
> Would it have sense to have such "feature"?

I'm not sure if you realize this or not, but the code you modified is what formats the header, it's not where normal parsing happens. It's kinda hard to explain, but if you are using something like MimeMessage.Load(), it won't hit that code path.

But to answer your question of "does it make sense to have?", I would need to better understand your scenario.

I'm not sure if this code change actually even fixes the issue you are having based on your description, but even if it does, I would not want to use RfcComplianceMode.Looser to cause this behavior because using EncodeUnstructuredHeader() is not going to do the right thing for address headers.

> We have some memory problems when processing TO header with 80k addressees.

Can you explain the situation to me a bit more? Maybe I can help you find a better solution (either a code modification in your code or a new feature in MimeKit).

A sample test case that illustrates the issue would be super helpful to me. Maybe I can use the test case to optimize address parsing or maybe the reformatting logic in order to reduce memory usage or whatever bottleneck is the problem.

The kind of information that I want to know is:

* Are you just using MimeMessage.Load() and this is where the issue is? If so, your change won't fix it.
* Are you setting `message.Headers["To"] = "really long string with 80k email addresses";`? If so, your change probably is "fixing" this, but it's not the right solution.
* Or is the problem when you do `message.WriteTo(formatOptions, stream);`? Your change might also be impacting this, but again, it's probably not the correct fix.

grzegorz-pasternak:
> * Are you just using MimeMessage.Load() and this is where the issue is? If so, your change won't fix it.
Yes, we are using MimeMessage.Load(Stream stream, true) to extract headers and text from body of the e-mail. 
> * Are you setting `message.Headers["To"] = "really long string with 80k email addresses";`? If so, your change probably is "fixing" this, but it's not the right solution.
No
> * Or is the problem when you do `message.WriteTo(formatOptions, stream);`? Your change might also be impacting this, but again, it's probably not the correct fix.
We additionally write zip file from the e-mail to temporary file stream but we observe problem rather with reading long header than zip file: 
                       ```
if (zipAttachment is MessagePart)
			{
				MessagePart rfc822 = (MessagePart)zipAttachment;
				rfc822.Message.WriteTo(tempFileStream);
			}
			else
			{
				MimePart part = (MimePart)zipAttachment;
				part.Content.DecodeTo(tempFileStream);
			}
```
We just want to read headers as they are without any parsing and in optimal way when it comes to memory. Thank you for your quick response.

rafal-bielinski:
I've created a bug for this: https://github.com/jstedfast/MimeKit/issues/919. Hope it explains the problem we are having...

PR #951, opened 2023-08-29
Modified Rfc2047's internal Token class to be a struct

jstedfast:
This reduces the number of allocations needed, but it does mean that List<Token> will be much larger.

To combat this a bit, redesigned the Token class/struct to have fewer fields which reduced sizeof(Token) from 40 bytes down to 24.

Ideally, we'd get that down even further to 16 bytes (as recommended by Microsoft). Added some comments with ideas on how to accomplish that.

### DIFF
--- a/MimeKit/Utils/Rfc2047.cs
+++ b/MimeKit/Utils/Rfc2047.cs
@@ -44,40 +44,68 @@ namespace MimeKit.Utils {
 		const int PhraseTokenCapacity = 16;
 		const int TextTokenCapacity = 32;
 
-		class Token {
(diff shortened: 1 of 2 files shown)


jstedfast:
I'm thinking that a better approach might be to have a Rfc2047TokenDecoder class/struct that is essentially a state machine that we can incrementally `Push(ref Token)` as we tokenize the input, thereby avoiding the need for a List or ValueList at all.

jstedfast:
This idea has been replaced by https://github.com/jstedfast/MimeKit/pull/952

PR #974, opened 2023-11-16
Add net8.0 to target frameworks

iamcarbon:


jstedfast:
Hah, I was just thinking of adding this over the weekend :)

PR #997, opened 2024-01-30
Improve Code Quality

iamcarbon:
This PR is scoped to the changes discussed in #996, including:

Use utf-8 literals to avoid encoding call and byte literals
Use statically initialized data
Use pattern matching in limited case
Removes unnecessary cast


iamcarbon:
@jstedfast Ready for review.

jstedfast (approved):

Issue #998, opened 2024-01-31
Add support for ECDH


myst3rium:
RSA is getting older and older and, in view of the massive increase in computing power, it is becoming increasingly vulnerable. On the other hand, there are the "new" elliptic curves, which offer secure encryption with significantly less effort.

I tried using MimeKit with certificates with ECC instead of RSA and dug deep into the library to see that the library doesn't seem to support ECC certificates; everything seems to be designed for RSA. So I keep getting the error "unkown cipher" when using an ECC certificate.
Maybe I just missed anything anywhere but the fact that the official API documentation (http://jstedfast.github.io/MimeKit/docs) is unreachable, too, doesn't make it easier to find a solution.

I would therefore be very grateful for at least some help and information in this area.

jstedfast:
> Maybe I just missed anything anywhere but the fact that the official API documentation (http://jstedfast.github.io/MimeKit/docs) is unreachable, too, doesn't make it easier to find a solution.

You just have bad timing - I'm in the process of moving the website from Azure to GitHub and unfortunately, things are not going smoothly for me 😦 

That said, you can still use the .com and .org website URLs to get to the old Azure site:

http://www.mimekit.org/docs
http://www.mimekit.com/docs

Now on to your actual question...

> I tried using MimeKit with certificates with ECC instead of RSA and dug deep into the library to see that the library doesn't seem to support ECC certificates; everything seems to be designed for RSA. So I keep getting the error "unkown cipher" when using an ECC certificate.

Where exactly are you getting this exception?

I have a sneaking suspicion (though I could be wrong) that you are trying to convert between System.Security certificates/keys and BouncyCastle certificates/keys. If so, then you are correct, there is no code that can handle ECC private key conversion from System.Security to BouncyCastle. Conversion between the 2 APIs has not been fun, even with better supported algorithms in System.Security such as RSA and DSA.

If you stick strictly with System.Security or BouncyCastle, things are more likely to work but there could be issues but I'm not 100% positive - this is why I am asking for the location in the code that threw the exception.

myst3rium:
Oh thank god! The documentation is back and there is a chance of getting some information <3

Well - let's start:
I had to derive a SecureMimeContext from your BouncyCastleSecureMimeContext to handle the certificatesa as we need it. There my entrance is the function "Encrypt" returning an ApplicationPkcs7Mime. On reaching the Call "CryptServices.Envelope(...)" the error returned is
[...]

PR #1003, opened 2024-02-08
Potentially fix CMS Data Envelope creation for ECC certificates

cipherboy:
See comment here for information: https://github.com/bcgit/bc-csharp/issues/516#issuecomment-1934781810 

Related: #998

### DIFF
--- a/MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
+++ b/MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
@@ -1339,7 +1339,8 @@ namespace MimeKit.Cryptography {
 							);
 						}
 					} else {
-						throw new ArgumentException("Unknown type of recipient certificate: " + pub.GetType().Name);
+						var oid = cert.SubjectPublicKeyInfo.Algorithm.Algorithm.ToString();
+						throw new ArgumentException("Unknown type of recipient certificate: " + pub.GetType().Name + " (SubjectPublicKeyInfo OID = " + oid + ")");
 					}
 					count++;
 				}
--- a/MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
+++ b/MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
@@ -1225,11 +1225,11 @@ namespace MimeKit.Cryptography {
 			}
 		}
 
-		class CmsRecipientInfoGenerator : RecipientInfoGenerator
+		class OaepAwareRecipientInfoGenerator : RecipientInfoGenerator
 		{
 			readonly CmsRecipient recipient;
 
-			public CmsRecipientInfoGenerator (CmsRecipient recipient)
+			public OaepAwareRecipientInfoGenerator (CmsRecipient recipient)
 			{
 				this.recipient = recipient;
 			}
@@ -1268,6 +1268,9 @@ namespace MimeKit.Cryptography {
 				var publicKeyInfo = certificate.SubjectPublicKeyInfo;
 				AlgorithmIdentifier keyEncryptionAlgorithm;
 
+				// If the recipient explicitly opts in to OAEP encryption (even if
+				// the underlying certificate is not tagged with an OAEP OID),
+				// choose OAEP instead.
 				if (publicKey is RsaKeyParameters && recipient.RsaEncryptionPadding?.Scheme == RsaEncryptionPaddingScheme.Oaep) {
 					keyEncryptionAlgorithm = recipient.RsaEncryptionPadding.GetAlgorithmIdentifier ();
 				} else {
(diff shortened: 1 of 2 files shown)


MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
```diff
@@ -1225,73 +1225,6 @@ public override Stream Verify (Stream signedData, out DigitalSignatureCollection
...
-
-			public RecipientInfo Generate (KeyParameter contentEncryptionKey, SecureRandom random)
-			{
-				var tbs = Asn1Object.FromByteArray (recipient.Certificate.GetTbsCertificate ());
-				var certificate = TbsCertificateStructure.GetInstance (tbs);
-				var publicKey = recipient.Certificate.GetPublicKey ();
-				var publicKeyInfo = certificate.SubjectPublicKeyInfo;
-				AlgorithmIdentifier keyEncryptionAlgorithm;
-
-				if (publicKey is RsaKeyParameters && recipient.RsaEncryptionPadding?.Scheme == RsaEncryptionPaddingScheme.Oaep) {
```
cipherboy, line 1271: It looks like the BC implementation should handle this for you just fine:

https://github.com/bcgit/bc-csharp/blame/7e3362065e144c4e211eea30f237b106183d70b4/crypto/src/cms/KeyTransRecipientInformation.cs#L67-L74

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
```diff
@@ -1300,7 +1233,27 @@ Stream Envelope (CmsRecipientCollection recipients, Stream content, Cancellation
...
+						// TODO: better handle algorithm selection.
+						cms.AddKeyAgreementRecipient (
+							CmsEnvelopedDataGenerator.ECDHSha1Kdf,
+							kp.Public,
+							kp.Private,
+							cert,
+							CmsEnvelopedGenerator.Aes128Wrap
+						);
+					} else {
+						throw new ArgumentException("Unknown type of recipient certificate: " + pub.GetType().Name);
```
cipherboy, line 1255: With a caveat I'm not a native C# developer... and I'm running on Linux... :-D 

[`3e5ab5372`](https://github.com/bcgit/bc-csharp/commit/3e5ab5372) introduced SPKI to X509Certificate and [`c806df63e`](https://github.com/bcgit/bc-csharp/commit/c806df63e) adds some helpers for fetching Algorithm information... I'd probably prefer to surface the OID here rather (or in addition to?) than the BC class name of the key type, as the OID is what the callers of the certificate will have.

jstedfast: You mean in the exception? Yea, that probably makes sense.

cipherboy: Yeah, in the exception. To do so, I think you need to update BC to a newer version (2.2.1, 2.3.0 is just out which should have this), but not sure how you'd do that from Linux and if you were avoiding it for some compatibility reasons. :-)

cipherboy: The other commentary here is that this doesn't handle other types of certs. PQC is coming, and with it brings questions of how to handle their various cert types in places like S/MIME and elsewhere. Personally if let the ink dry on the standards before worrying about that too much here, but IMO, this should set you up a little better for that when the time comes.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
```diff
@@ -1300,7 +1233,27 @@ Stream Envelope (CmsRecipientCollection recipients, Stream content, Cancellation
 
 			foreach (var recipient in recipients) {
 				if (unique.Add (recipient.Certificate)) {
-					cms.AddRecipientInfoGenerator (new CmsRecipientInfoGenerator (recipient));
+					var cert = recipient.Certificate;
+					var pub = recipient.Certificate.GetPublicKey();
+					if (pub is RsaKeyParameters) {
+						cms.AddKeyTransRecipient (cert);
```
jstedfast, line 1239: I think the reason I had the CmsRecipientInfoGenerator originally (and I'll have to verify by reviewing commit logs) was because I needed to be able to:

1. allow users to specify SubjectKeyIdentifier vs IssuerAndSerialNumber (the 2 possible values of the CmsRecipient.RecipientIdentifierType property)
2. allow users to override the CmsRecipient.RsaEncryptionPadding

CmsEnvelopeInfoGenerator.AddKeyTransRecipient() didn't appear to allow that from what I could tell.

I believe that's why I had to do my own thing.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
jstedfast: Right, so it uses the certificate to determine PKCS#1v1.5 vs OAEP as opposed to allowing users to override it.

We need to be able to override that.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
jstedfast: > To do so, I think you need to update BC to a newer version (I believe 2.2.1 is used in the repo, 2.3.0 is just out which should have this)

I have no problem with bumping BC to 2.3.0 (was planning on doing it anyway). You can actually just edit the MimeKit.csproj and change the version string for BouncyCastle.Cryptography and then run `nuget restore MimeKit.sln`  and/or `dotnet restore MimeKit.sln` and it will download/unpack the version you set in the csproj.

I think `nuget restore` is for .NETFramework and `dotnet restore` works for dotnet core projects. MimeKit is both, so not sure if you need to run both or if one is enough (and if so, which one haha).

Just a useful tip.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
cipherboy: Ahh yes, the algorithm that keeps on giving no end of incompatibilities. I'll need to think about that more.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
cipherboy: The first, a new BC API call to specify the `RecipientInfo` might be nice in that case.

cipherboy: @jstedfast Aha! So I took a second, closer look at the APIs. I think we can already do this without needing the hook! There's two APIs:

```c#
public void AddKeyTransRecipient(X509Certificate cert) -> KeyTransRecipientInfoGenerator -> IssuerAndSerialNumber(recipCert.CertificateStructure)
```

vs

```c#
public void AddKeyTransRecipient(AsymmetricKeyParameter pubKey, byte[] subKeyId) -> KeyTransRecipientInfoGenerator -> byte[] subjectKeyID, IKeyWrapper keyWrapper
```

I think these do what you want with RecipientInfo construction. The same applies to the two forms of `AddKeyAgreementRecipient` as well :-) 


So back to looking for a viable OAEP approach.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
cipherboy: Perfect, thank you! I've pushed a commit with this information in the error and updating BC to 2.3.0.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
cipherboy: I see... The linked class above is for recipients (decoding this message), not used during generating a message. I don't see an easy way around it, so I'll propose a PR back to bc-csharp.

cipherboy:
@jstedfast I think this version should retain compatibility with the existing OAEP selection code, but let you do ECC perhaps. I'll need a PR for BC-CSharp to do OAEP nicely, but you'll probably have to wait a bit to get a new release of that cut, so I'd probably suggest merging this as-is. :-)

jstedfast:
@cipherboy thanks a ton for your help, this is awesome. Give me a ping if you ever add that nicer way to do OAEP nicely, but if not, no worries - I don't mind having code that does a bit of extra work to massage things.

MimeKit/Cryptography/BouncyCastleSecureMimeContext.cs
cipherboy: @jstedfast If all goes well, this should show up in `89e04f3a` in bc-csharp and the next release. :-)

PR #1030, opened 2024-05-10
DSAParameters.Y should be padded

peterdettman:
AsymmetricAlgorithmExtensionTests.TestDSACryptoServiceProvider can fail occasionally if the Y parameter happens to have enough leading zero bits.


jstedfast:
Awesome, yes, I have had occasional failures in this test.

PR #1038, opened 2024-05-27
Improve several BC usages

peterdettman:
- use Asn1.Cms.Time for SigningTime attribute
- use X509Certificate.TbsCertificate property (instead of recoding)
- cleanup some obsolete TODOs


jstedfast:
Awesome, thanks Peter!

PR #1048, opened 2024-06-19
Add MySql support to the Certificate DB

rblackin:
Adding support for the Certificate tables to be used on MySql.

### DIFF
--- a/MimeKit/Cryptography/MySqlCertificateDatabase.cs
+++ b/MimeKit/Cryptography/MySqlCertificateDatabase.cs
@@ -1,0 +1,346 @@
+//
+// MySqlCertificateDatabase.cs
+//
+// Author: Rob Blackin <rob@5-9z.com>
+//
+// Copyright (c) 2013-2024 .NET Foundation and Contributors
+//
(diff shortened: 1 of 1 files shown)


jstedfast:
This looks fine to me, but w/o unit tests I can't verify it.

You'll note that MimeKit.csproj has the following:

```xml
<Compile Remove="Cryptography\NpgsqlCertificateDatabase.cs" />
```

The reason for this, is again, the inability to write unit tests for PostgresSQL.

I would be happy to accept this PR, but I will likely need to remove it from the default build in the same way that Npgsql is being removed.

jstedfast:
@rblackin any thoughts/preferences? Also, if you know of a good way to add unit tests that would be able to setup/connect to/make use of a MySQL database, I'd love to know because if that's something that you might know how to get working, then it would be the ideal situation because I would actually love to include this.

jstedfast:
Helpful hint to myself later: maybe this guide would be useful? https://medium.com/@michaelmatere606/configuring-mysql-for-github-actions-a-step-by-step-guide-f49e0a7aeedc

PR #1060, opened 2024-08-11
WIP: `FormatOptions`: Feature `ConvertPunycodeToIdn`

NicolaiSoeborg:
**Is your feature request related to a problem? Please describe.**
A lot of places requires working with punycoded mails as the IDN form is not handled correctly.
But when parsing, mails are turned into IDN form, i.e. `MimeKit.MailboxAddress.Parse("X@xn--sb-lka.org").Address == "X@søb.org"`

**Describe the solution you'd like**
It would be nice with an option to keep the address in punycoded form.

**Describe alternatives you've considered**
I guess users of this library can also do something like:
```C#
if (mailbox.IsInternational)
    mail = mailbox.LocalPart + "@" + MailboxAddress.IdnMapping.Encode(mailbox.Domain)
```

**Additional context**
This is a suggestion for adding that. Comments are welcome!

### DIFF
--- a/MimeKit/FormatOptions.cs
+++ b/MimeKit/FormatOptions.cs
@@ -83,6 +83,7 @@ namespace MimeKit {
 		bool verifyingSignature;
 		bool ensureNewLine;
 		bool international;
+		bool convertPunycodeToIdn;
 		int maxLineLength;
 
 		/// <summary>
@@ -243,6 +244,15 @@ namespace MimeKit {
 		}
 
 		/// <summary>
+		/// Get or set whether to convert punycode domains.
+		/// </summary>
+		/// <value><c>true</c> if punycoded domains should be converted to IDN; otherwise, <c>false</c>.</value>
+		public bool ConvertPunycodeToIdn {
+			get { return convertPunycodeToIdn; }
+			set { convertPunycodeToIdn = value; }
+		}
+
+		/// <summary>
 		/// Get or set whether the formatter should allow mixed charsets in the headers.
 		/// </summary>
 		/// <remarks>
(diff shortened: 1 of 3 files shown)


jstedfast:
`FormatOptions.International` is more-or-less meant for this already.

> I guess users of this library can also do something like:
> 
> ```csharp
> if (mailbox.IsInternational)
>     mail = mailbox.LocalPart + "@" + MailboxAddress.IdnMapping.Encode(mailbox.Domain)
> ```

I think `MailboxAddress.IdnMapping` is `internal` (or should be if it's not), but you can do this instead:

```csharp
var email = MailboxAddress.DecodeAddrspec (mailbox.Address);
```

That said, perhaps the parser shouldn't decode International Domain Names and/or perhaps a `InternationalAddress` property could be added and `Address` could enforce a punycode encoded address?

The only issue with that is it would break existing expectations, so perhaps a better approach would be to add `LegacyAddress` or `SafeAddress` or something (not sure I like either property name, but you get the idea).

Before I jump out of the starting gate, though, what scenarios do you need supported, exactly?

- [ ] In other words, do you need to enforce punycode encoding for mailbox addresses when writing the MimeMessage back out to a stream (e.g. to disk or when sending the MimeMessage via SMTP)?
- [ ] Do you need this to happen even if/when `FormatOptions.International` is set to `true`?
- [ ] Is it necessary for `MailboxAddress.Address` (or a replacement property) to be punycode encoded for convenience?
- [ ] Do you need `mailbox.ToString(FormatOptions)` to output the punycode encoded address?
- [ ] Other scenarios that I've missed.

Sounds like you care about option 3, but what about the others?

Alternatively, is `MailboxAddress.DecodeAddrspec (mailbox.Address)` enough now that you know about that option?

jstedfast:
BTW, if you only ever want punycode-encoded domains, another option is to swap out the `MailboxAddress.IdnMapping` implementation to one that doesn't encode/decode (or one that always converts into the encoded version).

You just need to implement the `IPunycode` interface: https://mimekit.net/docs/html/T_MimeKit_Encodings_IPunycode.htm

jstedfast:
I'm going to close this because I haven't heard any response, but if there is still interest in this we can reopen and have more discussion on how best to meet your needs.

PR #1082, opened 2024-10-02
docs: remove .NET forum discussion link as it does not exist anymore

rklec:
I found no replacement. https://dotnetfoundation.org/community/ may be the best equivalent, but even there no real community forum can be found anymore?

Maybe they now use Discord or what? But that is not the same, IMHO. And not worth promoting in any case.


jstedfast (approved):

jstedfast:
Thanks!

Issue #938, opened 2023-07-05
Suppressing Automatic Replies


mayka-mack:
If one wanted to suppress receiving automatic replies in response to a message sent via MailKit, would adding the [X-Auto-Response-Suppress](https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxcmail/ced68690-498a-4567-9d14-5c01f974d8b1) header to the MimeMessage be the correct way to go about this? Our organization uses Outlook 365, so the vast majority of our auto-replies will be coming from an Exchange server.

```
MimeMessage message = new();
message.Headers.Add("X-Auto-Response-Suppress", "DR, RN, NRN, OOF, AutoReply");
```

There is of course the more standard "Auto-Submitted: auto-generated" header, but I was unsure if this would also tell the mail server not to respond with another auto-generated email or auto-reply.
```
message.Headers.Add(HeaderId.AutoSubmitted, "auto-generated");
```

Sicos1977:
It is almost impossible to suppress auto replies on e-mails that you sent. You can set the ReplyTo field to something like noreply@yourcompany.com to let the receiving mailserver sent the reply to another e-mail address that just eats the reply. But even that is not 100% full proof. 

You can also make something that just tries to detect if an e-mail is an auto reply and then just deletes the e-mail but that is also not 100% full proof because there is no real standard that says if an e-mail is an auto reply.

Sicos1977:
I made this to try to detect auto replies. If you want you can use it because all the code in it comes mostly from the internet

You probably need to translate my Dutch (Netherlands) comments.

```c#
internal class AutoReplyDetector
{
    #region Consts
    private const string AutoSubmittedHeader = "Auto-submitted";
    private const string XAutoreplyHeader = "X-Autoreply";
    private const string XAutoRespond = "X-Autorespond";
    private const string PrecedenceHeader = "Precedence";
    // private const string XMailerHeader = "X-Mailer";
    private const string XAutoResponseSuppressHeader = "X-Auto-Response-Suppress";
    private const string ListIdHeader = "List-Id";
    private const string ListUnsubscribeHeader = "List-Unsubscribe";
    private const string FeedbackIdHeader = "Feedback-ID";
    private const string XmsFBLHeader = "X-MSFBL";
    private const string XLoopHeader = "X-Loop";
    #endregion

    #region Private enum XAutoResponseSuppressHeaderValue
    /// <summary>
[...]

PR #1083, opened 2024-10-02
Update HeaderId to clarify the difference between Autosubmitted and Auto-Submitted

rklec:
This was totally confusing for me why there are two header IDs in MimeKit with the quite very same header. I even suspected a bug or that it is just a legacy API-compatibility thing (one being deprecated, though not stated?).

After some research it turns out, the header is actually defined. At least for the modern RFC3834 header the keywords are actually also defined by IANA: https://www.iana.org/assignments/auto-submitted-keywords/auto-submitted-keywords.xml (But you don't do mapping to an enum, so introducing this would change the API drastically and be a different/bigger task, so I thought for the code this may not be relevant.)

So to clarify this and ease searching around the web and decrease potential confusion for other developers, IMHO, it is a good idea to just directly point out the basic difference in the (code) doc, directly where you are using it.

For background, I am at https://github.com/jstedfast/MimeKit/issues/938#issuecomment-2388847922 here detecting auto-reply mails for which this header is very important.

### DIFF
--- a/MimeKit/HeaderId.cs
+++ b/MimeKit/HeaderId.cs
@@ -122,7 +122,7 @@ namespace MimeKit {
 		/// The Auto-Submitted header field.
 		/// </summary>
 		/// <remarks>
-		/// The header as defined in <see href="https://www.rfc-editor.org/rfc/rfc3834">RFC3834</see>
+		/// The header as defined in <see href="https://www.rfc-editor.org/rfc/rfc3834">RFC3834</see>.
 		/// </remarks>
 		AutoSubmitted,
 
--- a/MimeKit/HeaderId.cs
+++ b/MimeKit/HeaderId.cs
@@ -130,8 +130,10 @@ namespace MimeKit {
 		/// The Autosubmitted header field.
 		/// </summary>
 		/// <remarks>
-		/// <b>Important:</b> This is more a legacy header as defined in <see href="https://datatracker.ietf.org/doc/html/rfc2156"/>RFC2156</see>. <br>
-		/// Usually, you likely want to use <see cref="AutoSubmitted"/> instead.
+		/// <note type="warning">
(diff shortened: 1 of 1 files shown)


MimeKit/HeaderId.cs
```diff
@@ -121,11 +121,18 @@ public enum HeaderId {
 		/// <summary>
 		/// The Auto-Submitted header field.
 		/// </summary>
+		/// <remarks>
+		/// The header as defined in <see href="https://www.rfc-editor.org/rfc/rfc3834">RFC3834</see>
```
rklec, line 125:
```suggestion
		/// The header as defined in <see href="https://www.rfc-editor.org/rfc/rfc3834">RFC3834</see>.
```

MimeKit/HeaderId.cs
```diff
@@ -121,11 +121,20 @@ public enum HeaderId {
...
+		/// The header as defined in <see href="https://www.rfc-editor.org/rfc/rfc3834">RFC3834</see>.
+		/// </remarks>
 		AutoSubmitted,
 
 		/// <summary>
 		/// The Autosubmitted header field.
 		/// </summary>
+		/// <remarks>
+		/// <note type="warning">
+		/// <para>This is a legacy header as defined in <see href="https://datatracker.ietf.org/doc/html/rfc2156"/>RFC2156</see>.</para>
```
rklec, line 134:
```diff
-	/// <para>This is a legacy header as defined in <see href="https://datatracker.ietf.org/doc/html/rfc2156"/>RFC2156</see>.</para>
+	/// <para>This is a legacy header as defined in <see href="https://datatracker.ietf.org/doc/html/rfc2156">RFC2156</see>.</para>
```

You've got a typo there @jstedfast

jstedfast: Thanks

PR #1113, opened 2024-12-07
Added support for domain-wide S/MIME certificates

jstedfast:
Replaces pr #1112


jstedfast:
TODO:

Add Unit Tests for the following scenarios:
1. Upgrading smime.db from the previous version (i.e. make sure adding SUBJECTDNSNAME column works as expected)
2. Lookups of domain S/MIME certificates works as expected for signing *and* encrypting

JoeShook:
I haven't tested this yet.  I will be putting time in on this weekend.  Because you are going to version the database I need to share with you one of the next PRs I would have started on.  

The domain-bound certificate is allowed to have more than one DNSName.  So we probably need to normalize the DB to handle the many DNSName to single certificate.

JoeShook:
And yes this PR works for the first DNSName.

jstedfast:
Ouch. Any thoughts on how to represent multiple DnsNames per certificate?

Initial thoughts are to serialuze the DnsNames using string.Join() and using a string contains match in the query? Is that doable?

JoeShook:
I think string concatination in this case would be fine rather than adding another table.

jstedfast:
I've updated this PR to use a concatenation of DnsNames. Can you verify that this works?

Also, if you could help me out with some Unit Tests that can sign/encrypt with domain-bound certificates, that would be super helpful as well. Or even just provide me with a sample certificate that I can use in the Unit Tests (preferably one that does not expire).

JoeShook:
I have been working on creating a domain-bound cert with openssl.  Have to wait to continue on this until later in the day.  

I didn't find any readme about your setup.  

[I created PR 1114](https://github.com/jstedfast/MimeKit/pull/1114) to share with you what I have so far.

We might as well create a cert to test punycode also.

jstedfast:
If you look in the [UnitTests/TestData/smime](https://github.com/jstedfast/MimeKit/tree/master/UnitTests/TestData/smime) directory, you'll find some \*.cfg files and a [genkeys.sh](https://github.com/jstedfast/MimeKit/blob/master/UnitTests/TestData/smime/genkeys.sh) file that I use to generate the other certificates.

Maybe you'll find those helpful.

jstedfast:
Forgot those .cfg files were for my own c#-based certificate generator that exists in the unit tests

Added the necessary hooks to generate a certificate with alt names.

jstedfast:
Okay, added unit tests and had to fix WindowsSecureMimeContext as well.

JoeShook:
I ran all my local test successfully.  I have more PRs to create, but this PR is looking good.


Out of curiosity, what OpenSSL commands are you using?
I assume the OPENSSL_CONF env is set to the ca.cnf.

Then you run

```openssl req -new -key smime.key -out smime.csr -config smime.cfg```

I generate a certificate from the pfx so I can then run the following.  But I imagine you have a clever way to use the pfx on one command.

```openssl x509 -req -in smime.csr -CA ../intermediate2.crt -CAkey ../intermediate2.key -CAcreateserial -out smime.crt -days 3650 -extfile smime.cfg -extensions v3_ca```

JoeShook:
Also in one of my future PRs I want to add CRLs.  So the end certs and intermediate would have CRLs.  Also the CRLs should be resolved during signing and signature validation ensuring trust.  For that matter it would be good to resovle the intermediate certificates also.   Here are my future PRs.  I have code for most of this readyish.

- What are your thoughts on generating CRLs?  To really test it, you almost need to generate the CRLs just in time so you can simulate CRL expiration and resolving the new version.  Caching...  Or just create a CRL that expires in 10 years.
- And would you create a in memory TestServer to simulate the static web server hosting the CRLs and intermediates?

jstedfast:
I'd probably implement a CRL generator in the unit tests just like I did for certificates.

Then generate a CRL on the fly to verify that it works.

JoeShook:
Oh, yes.  I see X509CertificateGenerator and I have seen that before.

jstedfast:
Yea, OpenSSL is complicated to figure out.

PR #1116, opened 2024-12-10
Enable certificate revocation

JoeShook:
- Created CRLs in the X509CertificateGenerator.  
- IsRevocationEnabled  is set by the Context CheckCertificateRevocation property rather than set to false in the BouncyCastleSecureMimeContext.
- Tests updated
- Negative tests added in ApplicationPkcs7MimeTests
- In SecureMimeTests CheckCertificateRevocation  was set to false.  

- [ ] May need to add a test with a revoked certificate.
- [x] Have a different PR in the pipeline that write the crl distribution point to the certificates.  Was waiting on this because I wanted to ensure I could test the resolution (download) of CRLs in unit tests.  I was just focusing on the CRL creation and manual configuration.  I could bring the part that generates these distribution points.


jstedfast:
I'll try to look at this tomorrow

MimeKit/Cryptography/AsymmetricAlgorithmExtensions.cs
```diff
@@ -247,8 +246,15 @@ static AsymmetricAlgorithm GetAsymmetricAlgorithm (DsaPrivateKeyParameters key,
...
-			if (pub != null)
+			if (pub != null) {
 				parameters.Y = GetPaddedByteArray (pub.Y, parameters.P.Length);
+			
+			}
+			else
+			{
+				// If pub is null, derive Y from the private key parameters
+				parameters.Y = key.Parameters.G.ModPow(key.X, key.Parameters.P).ToByteArrayUnsigned ();
+			}
```
JoeShook, line 257: This change was initiated when I saw this Test failure in the build logs:
https://github.com/jstedfast/MimeKit/actions/runs/12252099258/job/34178015357#annotation:11:73

```
1) Error : UnitTests.Cryptography.AsymmetricAlgorithmExtensionTests.TestDSACryptoServiceProvider
System.ArgumentException : The specified DSA parameters are not valid; P, Q, G and Y are all required.
   at System.Security.Cryptography.DSAOpenSsl.ImportParameters(DSAParameters parameters)
   at System.Security.Cryptography.DSACryptoServiceProvider.ImportParameters(DSAParameters parameters)
   at MimeKit.Cryptography.AsymmetricAlgorithmExtensions.GetAsymmetricAlgorithm(DsaPrivateKeyParameters key, DsaPublicKeyParameters pub) in /home/runner/work/MimeKit/MimeKit/MimeKit/Cryptography/AsymmetricAlgorithmExtensions.cs:line 255
   at MimeKit.Cryptography.AsymmetricAlgorithmExtensions.AsAsymmetricAlgorithm(AsymmetricKeyParameter key) in /home/runner/work/MimeKit/MimeKit/MimeKit/Cryptography/AsymmetricAlgorithmExtensions.cs:line 333
   at UnitTests.Cryptography.AsymmetricAlgorithmExtensionTests.AssertDSA(DSA dsa) in /home/runner/work/MimeKit/MimeKit/UnitTests/Cryptography/AsymmetricAlgorithmExtensionTests.cs:line 73
   at UnitTests.Cryptography.AsymmetricAlgorithmExtensionTests.TestDSACryptoServiceProvider() in /home/runner/work/MimeKit/MimeKit/UnitTests/Cryptography/AsymmetricAlgorithmExtensionTests.cs:line 125
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

jstedfast: Yea, that's been an intermittent failure for years - I just hadn't taken the time to debug it. So thanks for taking the time :-)

UnitTests/Cryptography/X509CertificateGenerator.cs
```diff
@@ -471,6 +486,27 @@ public static X509Certificate[] Generate (GeneratorOptions options, PrivateKeyOp
 			using (var stream = File.Create (options.Output))
 				pkcs12.Save (stream, options.Password.ToCharArray (), random);
 
+
+			var crlGenerator = new X509V2CrlGenerator ();
+			var intermediateNow = DateTime.UtcNow;
+			crlGenerator.SetIssuerDN (certificate.IssuerDN);
+			crlGenerator.SetThisUpdate (intermediateNow);
+			crlGenerator.SetNextUpdate (intermediateNow.AddDays (crlOptions.DaysValid));
```
jstedfast, line 494: The problem I see is that we're generating these CRLs if they don't already exist and we are setting it to expire after 7 days.

That means, every 7 days or so, I will need to delete these CRLs and run the unit tests to generate a new batch of them.

The reason the Certificates are done that way is because I used to use OpenSSL commands to generate them locally, check them into the repo, and then be able to run the unit tests again.

When I wrote C# code to generate them instead, I probably should have abandoned the strategy of checking them into the repo, but with a 10-year lifespan, it's not a big deal and also means that I can more easily debug an issue that pops up in the CI builds (reproducible certificates).

The same could be argued for CRLs, I suppose, but 7 days is really short and it won't *really* test anything (or am I missing something?).

The current tests only really test importing CRLs (which is something we need to test, but I'm not sure if it makes sense to check them into the repo).

For *actually* testing certificate revocation works if a CRL is available for a particular certificate, what we really need to be able to do is generate one on-the-fly. And then what we would do is have the following tests:

1. Check that ApplicationPkcs7Mime.Verify() fails (for the correct reason) if a CRL is available for:
    a. the signer's certificate
    b. the intermediate certificate
    c. the CA certificate(?)
2. Check that MultipartSigned.Verify() (for the correct reason) if a CRL is available for:
    a. the signer's certificate
    b. the intermediate certificate
    c. the CA certificate(?)
3. Check that ApplicationPkcs7Mime.Encrypt()/SignAndEncrypt() (for the correct reason) if a CRL is available for:
    a. a recipient's certificate
    b. the intermediate certificate of a recipient's certificate
    c. the CA certificate of a recipient's certificate

I'm not sure what should happen if we try to Sign() with a revoked certificate, but arguably we should allow it and leave it up to a caller to make the decision not to. If we do this, we should probably have a public method on SecureMimeContext that can check revocation status for a particular certificate.

JosephEShook:
I will open a new PR for the DSA fix

UnitTests/Cryptography/X509CertificateGenerator.cs
JoeShook: There is a lot here to work through.  I am not in a rush.  

The 7-day choice was just an arbitrary decision. In other projects, I've set it to 10 years to avoid dealing with it. I also have automated builds that produce all the PKI without checking into source. I'm flexible on this. The rollover style integration tests are interesting, but I wasn't trying to solve that in this PR. Funny enough, the first time I saw those .cfg files, I thought they were OpenSSL config files.

I think you're leaning towards generating the CRLs on the fly. I'll update the code accordingly.

>The current tests only really test importing CRLs (which is something we need to test, but I'm not sure if it makes sense to check them into the repo).

I've run this code against my test PKI with CRL Endpoints defined. If I don't load the CRLs, the chain build fails as expected, similar to using X509Chain.Build from System.Security.Cryptography. I didn't include that code in this PR because I thought it would be better suited for another PR focused on CRL and Intermediate resolution. I'll bring that in here if the tests keep passing to add to the conversation. I'm still waiting on the resolution process for a future PR.

To repeat myself in different words, If a certificate in the chain has a CRL or OCSP endpoint extension and CheckCertificateRevocation = true but the CRL isn't loaded, the x509Chain Build process will always fail.   This is what I have tested with my PKI.  So, the following comment might not be optional.  The user can set CheckCertificateRevocation  = false.  Before this PR, IsRevocationEnabled was hard coded to false in two places  and none of the tests certs had a CRLDistributionPoint to trigger this failure.

> 1. Check that ApplicationPkcs7Mime.Verify() fails (for the correct reason) if a CRL is available for:
>     a. the signer's certificate
>     b. the intermediate certificate
>     a. the CA certificate(?)
>...

I think creating a revoced cert is another missing item for testing.

Loading a new CRL while holding an expired one is also good but can be included in the CRL Resolution PR.

jstedfast: > To repeat myself in different words, If a certificate in the chain has a CRL or OCSP endpoint extension and CheckCertificateRevocation = true but the CRL isn't loaded, the x509Chain Build process will always fail.

D'oh, now I get it. Thanks for the explanation. If my memory is correct, I think I ran into this issue in testing years ago.

I'm really happy to have your help with this because it's not my area of expertise.

> I think creating a revoced cert is another missing item for testing.

Er, yes, in my brain-fart comment above, what I had been intending to say is that we should check those scenarios for a certificate that has been revoked which we would know about based on the CRL that the tests will dynamically generate to let us know that the certificate has been revoked (at least I think that's how CRL's work).

I would actually love to work with you on this functionality. Would it make sense for me to create a feature branch?

jstedfast:
FWIW, I cherry-picked your DSA fix from your fork. You don't need to create a PR anymore.

jstedfast:
I also cherry-picked https://github.com/jstedfast/MimeKit/pull/1116/commits/de80777c313c1f204939eacc850207c073c5aabe since that could be applied stand-alone without breaking anything

UnitTests/Cryptography/X509CertificateGenerator.cs
JoeShook: Yes, a feature branch sounds like a good way to go. I assume it will include everything I've hinted at, including the CRL and intermediate resolutions. I wonder if we should also change BouncyCastleSecureMimeContext.BuildCertificateChain to virtual. This is how I approached it the first time I started testing with my PKI.  For example an implementor could choose the System.Security.Cryptography code to build the chain.  

FYI, The MS stack resolves the Intermediates and CRLs for you.  I think the only place it doesn't is on Android.  Which was surprising, I would have expected it to me similar to the Linux experience.  Also on Linux I would have to come up with a caching strategy for intermediates and crls.  But your pattern here in MimeKit is very adaptable to caching the Intermediates and CRLs.

jstedfast: I'll go create a feature branch.

jstedfast: branch: `smime-revocation-checks`

jstedfast:
I just hacked something up tonight. Let me know what you think.

PR #1144, opened 2025-02-22
Weak encryption: Insufficient key size

fgomes0000:
Changed the RSA and DSA key size to 2048 in the Unit Tests.


jstedfast:
Had to revert this as it broke the unit tests

jstedfast:
FWIW, DSACryptoServiceProvider is limited to FIPS-186-1 (which had a 1024-bit key limit).

PR #1157, opened 2025-04-21
Check if disposed in MemoryBlockStream.ToArray

daniel-lerch:
This pull request add a unit test which checks that `MemoryBlockStream` correctly throws `ObjectDisposedException`. This was not the case for `MemoryBlockStream.ToArray` which is added in the second commit.

I found this while working on #1067 but opened a separate PR because strictly speaking this a behavioral change in public API.


jstedfast (approved):
Awesome, thanks!

PR #1160, opened 2025-04-30
Check if disposed on FilteredStream filter manipulation

daniel-lerch:
This pull request add a unit test which checks that `FilteredStream` correctly throws `ObjectDisposedException`. This was not the case for `Add`, `Remove`, and `Contains` which is added in the second commit.

I was uncertain whether it would make sense to add a `leaveOpen` parameter here too. It has not been necessary so far and following the .NET pattern introducing a parameter `bool leaveOpen = false` would be potentially dangerous behavior change. Therefore, I decided to just a comment that the source stream won't be disposed.

### DIFF
--- a/MimeKit/IO/FilteredStream.cs
+++ b/MimeKit/IO/FilteredStream.cs
@@ -62,6 +62,7 @@ namespace MimeKit.IO {
 		/// </summary>
 		/// <remarks>
 		/// Creates a filtered stream using the specified source stream.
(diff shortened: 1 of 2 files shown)


jstedfast (approved):

daniel-lerch:
You merge pull requests faster than GitHub notifications take to update 😂🚀

jstedfast:
😂

PR #1183, opened 2025-08-11
Add more nullable annotations

daniel-lerch:
This pull request adds many more nullable annotations and enables nullable reference types for the entire `MimeKit` project.

There are 99 warnings left related to nullability. I find these hard to fix because they are related to fields that can be null but only under certain circumstances i.e. parser state, depending on the super class or which constructor was used. @jstedfast would you be willing to work on these remaining warnings?


jstedfast:
Figured I should post an update. I just spent the last week migrating off my old workstation onto a new work laptop and just bought myself a personal laptop to use for MimeKit development (since my PC is stuck on Windows 10 and it was time for an upgrade).

I'll try to get to this soon, but this weekend will likely be spent setting up my new personal laptop.

PR #1214, opened 2026-01-15
Base64.DecodeFromUtf8 bug with active hardware acceleration

tinohager:
The method returns corrupt attachments with active hardware acceleration.
I lose 2 bytes in my email attachment. When I set isFinalBlock to `true`, everything is fine.

**my check method** provides valid attachments.
```cs
using var attachmentStream = new MemoryStream();
mimePartAttachment.Content.Stream.CopyTo(attachmentStream);
var attachmentBytes = attachmentStream.ToArray();

string base64String = Encoding.UTF8.GetString(attachmentBytes);
byte[] originalBytes = Convert.FromBase64String(base64String);
```

**DecodeToAsync** provides corrupt attachments.
```cs
await mimePartAttachment.Content.DecodeToAsync(attachmentStream);
var data = attachmentStream.ToArray();
```

### DIFF
--- a/MimeKit/Encodings/Base64Decoder.cs
+++ b/MimeKit/Encodings/Base64Decoder.cs
@@ -267,7 +267,7 @@ namespace MimeKit.Encodings {
 			var utf8 = new ReadOnlySpan<byte> (inptr, remainingInput);
 			var decoded = new Span<byte> (outptr, remainingOutput);
 
-			var status = Base64.DecodeFromUtf8 (utf8, decoded, out int bytesRead, out int bytesWritten, false);
+			var status = Base64.DecodeFromUtf8 (utf8, decoded, out int bytesRead, out int bytesWritten, true);
 			outptr += bytesWritten;
 			inptr += bytesRead;
 


jstedfast:
Okay, so I've debugged this a bit and here's what seems to be happening:

Base64.DecodeFromUtf8() is returning `InvalidData` but bytesRead is 2 (and bytesWritten is 0).

This causes MimeKit to resume at "\r\nA=" where it skips the "\r\n" (correct) and consumes the "A=" but does not decode it because it is waiting for 2 more bytes (which if it were to make any assumptions, it would expect "==").

Not sure what is causing Base64.DecodeFromUtf8() to consume 2 bytes, but I would guess it is consuming the "AA" but I wasn't able to step into the code, so I'm not sure. Looking at the code, it could also be getting confused and "consuming" the "\r\n" in its fallback decoder logic (DecodeWithWhiteSpaceBlockWise) which copies data into a 4-byte block for decoding a block at a time. In that code, tracks "skipped" bytes (whitespace chars) and immediately adds the "skipped" value to "consumedBytes" before doing any decoding.

I don't see anything obviously wrong with the latest version of the code, but it's hard to diagnose this by looking at non-syntax-highlighted code in a browser window on GitHub and not being able to step through it.

Suffice it to say, I think I need to default this to disable hwaccel and go back to my slower (but more correct) decoder.

jstedfast:
I *think* this is a dotnet runtime bug (but I could be misunderstanding how to use the Base64.DecodeFromUtf8() API, I suppose).

The .NET Runtime devs might tell me "but it's the last block, so of course you need to pass `isFinalBlock:true`" but that makes using that API difficult because we can't *know* that it's the final block until there's no more data to stream. I would also argue that it's completely fair that it doesn't decode that block, but if it's not going to decode that block, don't set `bytesConsumed = 2` which makes it impossible for MimeKit to recover.

jstedfast:
It's been fun watching Copilot fix this bug.

Pretty impressive actually.

jstedfast:
I'm not sure, but yes, it seems tagged for 11.0 which may mean it'll only get fixed for net11.0 which is kinda sad.

This may be a preliminary milestone though. It's possible they might backport the fix.

jstedfast:
I'm pretty sure there's got to be other bugs in Base64.DecodeFromUtf8() because we had an issue on O365 Exchange very recently where a customer reported  images getting corrupted mid-image which can't possibly be because of this particular bug.

I got excited when you initially filed this bug report because I was hoping it would be the same issue we got there, but it doesn't seem to be.

PR #1249, opened 2026-06-24
InternetAddressListConverter: Convert an empty string to an empty InternetAddressList

0xced:
I think it makes sense to convert the empty string to an empty `InternetAddressList`. Also, this is really useful for configuration binding.

### DIFF
--- a/MimeKit/InternetAddressList.cs
+++ b/MimeKit/InternetAddressList.cs
@@ -55,6 +55,8 @@ namespace MimeKit {
 	[TypeConverter (typeof (InternetAddressListConverter))]
 	public class InternetAddressList : IList<InternetAddress>, IEquatable<InternetAddressList>, IComparable<InternetAddressList>
 	{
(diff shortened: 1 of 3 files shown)


MimeKit/InternetAddressList.cs
```diff
@@ -55,6 +55,8 @@ namespace MimeKit {
 	[TypeConverter (typeof (InternetAddressListConverter))]
 	public class InternetAddressList : IList<InternetAddress>, IEquatable<InternetAddressList>, IComparable<InternetAddressList>
 	{
+		internal static readonly InternetAddressList Empty = [];
+
```
jstedfast, line 59: The problem with this is that it is not guaranteed to *stay* empty. Once you return this to a caller, the caller can add mailboxes to it.

jstedfast:
I've applied a fixed fix

0xced:
For reference: 73e448f185bff08ebc33c49ced59df0d9afc0902

